PR #1, opened 2019-09-02
Initial revision

timg236:
Import the files used for the first beta release.


timg236:
@XECDesign - is this everything that you need for the APT package ?

andrum99:
These files are missing a licence. If you are intending these to be open-source, you would need to add a suitable licence somewhere, for example in a LICENCE file in the root of the repo.

Also, would you be interested in man pages for these scripts? I can write these.

timg236:
> These files are missing a licence. If you are intending these to be open-source, you would need to add a suitable licence somewhere, for example in a LICENCE file in the root of the repo.
> 
> Also, would you be interested in man pages for these scripts? I can write these.

A license is on the todo list. Happy to accept PR for man pages, although I'd recommend holding off until any feedback from the beta is merge in-case of interface changes. Hopefully, there won't be any though!

XECDesign:
We could generate the man pages with help2man.

I've attached the license in Debian format. Probably not suitable for the source repo, but you can copy the appropriate bits from there.
[copyright.txt](https://github.com/raspberrypi/rpi-eeprom/files/3565528/copyright.txt)

Everything is BSD-3 except for the EEPROM images themselves, which are restricted for use with Raspberry Pi products only.

rpi-eeprom-update
```diff
@@ -0,0 +1,396 @@
...
+
+   if [ "${USE_FLASHROM}" = 0 ]; then
+      [ -f "${RECOVERY_BIN}" ] || die "${RECOVERY_BIN} not found"
+   fi
+}
+
+usage() {
+cat <<EOF
+rpi-eeprom-update [options]... [FILE]
+   Checks whether there Raspberry Pi bootloader EEPROM is up to date and
```
andrum99: there -> the. Add hyphenation: up to date -> up-to-date

timg236: Will fix in next set of script updates, this is just importing the files that are already available

LICENSE
```diff
@@ -0,0 +1,67 @@
+Format: https://www.debian.org/doc/packaging-manuals/copyright-format/1.0/
+Upstream-Name: rpi-eeprom
+Source: https://github.com/raspberrypi/rpi-eeprom
+
+Files: *
+Copyright: 2019, Raspberry Pi (Trading) Ltd.
+License: BSD-3
+
+Files: lib/firmware/*
```
XECDesign: The path has changes since the file was written.

PR #7, opened 2019-09-10
Add pieeprom-2019-09-10.bin

timg236:
Minor update to always configure RGMII pins to a good state at power on.
Previously the pins would have been left in the default reset state
(input).


timg236:
@XECDesign @P33M 

I've updated the beta to include the RGMII change, it "works for me". Once the Debian package scripts are updated I think this can be merged for an end-to-end test before we encourage people  to use this firmware.

N.B. This will probably become the production image / latest critical EEPROM image unless any regressions are found with the SDRAM timings (very unlikely).

timg236:
merged so that we can integrate the package creation scripts and create a new release with the latest beta

PR #8, opened 2019-09-10
rpi-eeprom-update: -h should not require vcgencmd

timg236:
More the checkDependencies call to after the options parsing so that
-h can be run for help2man.


timg236:
@XECDesign rpi-eeprom-update -h should now run without vcgencmd

XECDesign:
Solves the problem for me, thanks.

timg236:
thanks @XECDesign

PR #9, opened 2019-09-10
Add debian packaging

XECDesign:


XECDesign:
Still need to test that the backup directory gets removed properly

XECDesign:
Tested. It does.

timg236:
Thanks. Looks good to me

PR #10, opened 2019-09-10
rpi-eeprom-update: minor fixes to usage, rename version check function

andrum99:


rpi-eeprom-update
```diff
@@ -193,22 +193,22 @@ checkDependencies() {
    fi
 
    if vcgencmd bootloader_config | grep -qi "Command not registered"; then
-      die "vcgencmd: bootloader_config. not supported. Please update VC firmware"
+      die "vcgencmd: 'bootloader_config` command not supported. Please update VC firmware"
```
XECDesign: Is it just me or do those single quotes not match? Shouldn't they just be normal `'` characters?

rpi-eeprom-update
```diff
@@ -230,9 +230,9 @@ rpi-eeprom-update [options]... [FILE]
    A backup of the current EEPROM config file is written to ${FIRMWARE_BACKUP_DIR}
    before applying the update.
 
-   -a Install the latest critical update if necessary.
-   -d Use the default bootloader config instead of migrating the current settings.
-   -f Install the given file instead of the latest critical update.
+   -a Install the latest update if necessary
+   -d Use the default bootloader config instead of migrating the current settings
+   -f Install the given file instead of the latest critical update
```
```suggestion
   -f Install the given file instead of the latest ${FIRMWARE_RELEASE_STATUS } update
```

andrum99: The usage message is used to build a man page when the package is installed, so this wouldn't work. I've changed 'critical' to 'applicable' instead.

timg236:
Thanks, I've given this a quick test and the updates look good

PR #15, opened 2019-09-18
Pieeprom 2019 09 10

timg236:


timg236:
@XECDesign tested on 1, 2, 4 GB variants, please can you create an updated apt package with the major version incremented ?

PR #20, opened 2019-09-24
Set unused data to zero if size of config is reduced

timg236:
If the size of bootconf.txt is reduced then set the unused data
to all ones instead of leaving garbage at the end. The EEPROM
header contains the actual file but this makes it easier to
verify the image and makes overreads more obvious.


timg236:
Needs some smoke testing but the fix + new test case should handle this corner case

timg236:
@XECDesign it  might be worth pulling in this minor change for the package which gets included in the next full Buster release.

PR #22, opened 2019-09-25
2019-09-25 - Beta test update for network boot

timg236:


firmware/raspberry_pi4_network_boot_beta.md
```diff
@@ -5,10 +5,8 @@ The page describes how to install and configure network boot for the Raspberry P
 
 This documentation will be merged into the main bootloader documents after the beta test is complete.
 
-## Known issue
-* For unknown reasons the brcmfmac wifi is not initialised in netowrk boot mode.
+## Known issues
 * Linux 5.3 kernel seems to poll SDHCI SD ever 10 seconds printing warnings to dmesg
```
MilhouseVH: 4.19.75, too.

timg236: Ok, I see it now. If I understand the driver quirks flag correctly then it continuously polls if SDHCI_QUIRK_BROKEN_CARD_DETECTION is set and there's a comment which says "assume card is always present". Unfortunately, the sdhci_timeout_timer always prints as pr_err. 
I think there needs to be some sort of patch which silences the pr_err or makes it a trace until the card has been detected.

PR #34, opened 2019-10-17
Vli update

timg236:


timg236:
@XECDesign @popcornmix @MilhouseVH @HiassofT

Sharing this before pushing to master in-case this impacts LibreElec. The significant change are that the recovery.bin will now update the VL805 if a VL805.bin file is supplied. It also requires sha256 hashes to be included for both the bootloader eeprom and VL805. 

Updating from the bootloader is safer than the VL805 tool because it can be done before the PCIe device bus mastering is enabled so there's no possibility of upsetting attached USB devices which can happen with the command line tool. I've update the USE_FLASHROM path to use the command line tool to support the idea of an immediate update (but didn't force rescan) although I wouldn't recommend using it except for expert users.

The hash is generated by the script and guards against file-system corruption, mostly this is is just paranoia for when 'rpi-eeprom' is included by default in Raspbian images and via "apt upgrade".

timg236:
If anyone is able to try this before we update the apt package then that would be very helpful. There's an install script to make this easier

git clone https://github.com/timg236/rpi-eeprom
cd rpi-eeprom
git checkout vli_update
# Update firmware and /usr/bin scripts plus select the BETA releases
sudo ./test/install -b

XECDesign:
Added to the internal apt repo

MilhouseVH:
@timg236 how difficult would it be to allow the auto option to selectively update bootloader and/or vl805? 

Currently `-a` will update both bootloader and vl805, and in LibreELEC I'd like to offer the user the option of updating to the "latest" bootloader and/or vl805.

As it stands, I'd need to call a combination of `-f` and/or `-u`, which will mean identifying the filenames of the latest firmwares (logic I'd hope to avoid).

Ideally it would be possible to call `rpi-eeprom-update -A bootloader -A vl805` (or similar) so that users could select which device is automatically updated.

Or perhaps the script can include the path of the "latest" firmware filenames in the machine output? In which case calling combinations of `-f`/`-u` is much less of an issue (I could live with this as a solution!)

timg236:
Assuming getopts gets invoked multiple times I think it would be fairly easy. I think there could be two boolean variables ENABLE_BOOTLOADER_UPDATES and ENABLE_VL805_UPDATES.
-a sets both to 1
-A sets the specific flag
These would be be anded with with ACTION_UPDATE_BOOTLOADER / ACTION_UPDATE_VL805 in applyUpdate.
I'm guessing you'd still want rpi-eeprom-update to always indicate whether a VL805 update is available (when called without args) ?

The -f and -u flags are really for beta testing or people using advanced boot modes who are updating the bootloader eeprom config and maybe saving a copy of their customized EEPROM.

MilhouseVH:
>  -A sets the specific flag
Thanks,, I'd be happy with that.

>  I'm guessing you'd still want rpi-eeprom-update to always indicate whether a VL805 update is available (when called without args) ?
Yes please! I'm only using the machine output, so when called without args the machine output should be populated with the details for both bootloader and vl805. Thanks.

By the way if you wanted to change any of the machine output variable names (for consistency, assuming it's not too late for any other users) then I don't have any issue with that (at this time).

andrum99:
I see a reference to a 'PXE boot menu option' in the changelog. Is a boot menu in development?

timg236:
> I see a reference to a 'PXE boot menu option' in the changelog. Is a boot menu in development?

This is just referring to the PXE option 9  in DHCP option 43 which returns 'Raspberry Pi Boot'. The Pi4 (and earlier) bootloaders use this as a hook to recognise that the server can be used for TFTP. So it's sort of abusing a bit of the PXE standard in order to discover TFTP services

timg236:
> > -A sets the specific flag
> 
> Thanks,, I'd be happy with that.
> 
> > I'm guessing you'd still want rpi-eeprom-update to always indicate whether a VL805 update is available (when called without args) ?
> 
> Yes please! I'm only using the machine output, so when called without args the machine output should be populated with the details for both bootloader and vl805. Thanks.

The VL805 output is only populated if run as root, unfortunately I don't think there is a non-root hook to read the version register from PCIe config space since it's after the first 64 bytes. In theory the bootloader could pass it to start.elf then make this available via a gencmd but that's quite combersome and it would be invalidated if the vl805 tool is run directly.

> By the way if you wanted to change any of the machine output variable names (for consistency, assuming it's not too late for any other users) then I don't have any issue with that (at this time).

What about the JSON output format should that change as well ?

I did wonder about changing CURRENT_TS,LATEST_TS to BOOTLOADER_CURRENT,BOOTLOADER_LATEST

MilhouseVH:
> The VL805 output is only populated if run as root,

LibreELEC only runs as `root`, at least for now (we do have plans to move away from `root` in future, but nothing is likely to happen any time soon) so that shouldn't be a problem. However, having some way of obtaining the vl805 version without `root` might be a nice idea, but probably not essential - there's always `sudo` for regular distros.

> What about the JSON output format should that change as well ?

Yes, keeping the JSON and non-JSON variable names in sync would be a good idea.

> I did wonder about changing CURRENT_TS,LATEST_TS to BOOTLOADER_CURRENT,BOOTLOADER_LATEST

I thought you might, that's why I mentioned it :smile: No problem from me, it's still early days which makes it easier to change at this time.

timg236:
@P33M

rpi-eeprom-update
```diff
@@ -192,12 +222,20 @@ checkDependencies() {
...
+   if ! vl805 -h > /dev/null 2>&1; then
+      die "vl805 command not found"
+   fi 
+
    if vcgencmd bootloader_config | grep -qi "Command not registered"; then
-      die "vcgencmd: 'bootloader_config' command not supported. Please update VC firmware"
+      die "vcgencmd: 'bootloader_config' command not supported. Please update VC firmware and reboot."
+   fi
+
+   if ! sha256sum --version > /dev/null 2>&1; then
```
MilhouseVH: Unfortunately busybox (which implements `sha256sum`) returns exit code=1 for `sha256sum --version`, so this test fails even when a working `sha256sum` is available.

Would it be possible to use instead:
```
if ! command -v sha256sum >/dev/null`; then
```
and the same for `vl805` and `flashrom`, as I think that might be more reliable.

If not `command`, then `which` should always be available:
```
  if ! which sha256sum >/dev/null; then
```

(From what I've read, `command` tends to be preferred to `which` these days).

rpi-eeprom-update
```diff
@@ -348,46 +469,66 @@ write_status_info()
...
+VL805_LATEST="${vl805_new}"
 EOF
    else
       cat > "${MACHINE_OUTPUT}" <<EOF
 {
   "EXITCODE": "${exit_code}",
-  "CURRENT_TS": ${cur},
-  "LATEST_TS": ${new}
+  "BOOTLOADER_CURRENT": ${bootloader_cur},
+  "BOOTLOADER_LATEST": ${bootloader_new}
```
```suggestion
  "BOOTLOADER_LATEST": ${bootloader_new},
```

rpi-eeprom-update
```diff
@@ -348,46 +469,66 @@ write_status_info()
    [ -z "${MACHINE_OUTPUT}" ] && return 0
 
    exit_code="${1}"
-   cur=${2}
-   new=${3}
+   bootloader_cur="${BOOTLOADER_CURRENT_VERSION}"
+   bootloader_new="${BOOTLOADER_LATEST_VERSION}"
```
```suggestion
   bootloader_new="${BOOTLOADER_UPDATE_VERSION}"
```

rpi-eeprom-update
```diff
@@ -348,46 +469,66 @@ write_status_info()
...
+   vl805_cur="${VL805_CURRENT_VERSION}"
+   vl805_new="${VL805_LATEST_VERSION}"
```
```suggestion
   vl805_new="${VL805_UPDATE_VERSION}"
```

MilhouseVH:
@timg236 many thanks for the -A option. I had to make a few changes (see above, or http://ix.io/1Zmk), and it now seems to be working as expected.

rpi-eeprom-update
timg236: Do you want all of the JSON values to default to zero instead of the empty string to ensure that the output is valid JSON ?

rpi-eeprom-update
timg236: done

rpi-eeprom-update
timg236: done

rpi-eeprom-update
MilhouseVH: If it's a string then an empty string (`""`) is fine, but for the integer time stamps JSON can't handle a missing value (as it was here, but probably because the wrong shell var had been used) so defaulting both to zero would be fine - I wasn't sure if `${bootloader_new}` could be blank (undefined). I suppose for consistency we should default both the JSON and non-JSON outputs to 0. Thanks!

MilhouseVH: This is probably the easier change:
```
-  bootloader_cur="${BOOTLOADER_CURRENT_VERSION}"
-  bootloader_new="${BOOTLOADER_UPDATE_VERSION}"
+  bootloader_cur="${BOOTLOADER_CURRENT_VERSION:-0}"
+  bootloader_new="${BOOTLOADER_UPDATE_VERSION:-0}"
```
then just add the missing comma to the JSON output.

MilhouseVH:
In the last commit a8a47df048a4bd285fd32b0f5a80b12b9c3078f1 you say that the exit code is an integer however when writing the status (-m) it's actually a string. The safest way to deal with a potentially undefined exit code is probably:
```
-  exit_code="${1}"
+  exit_code="${1:-EXIT_FAILED}"
```
although it's unlikely this would be required.

timg236:
Updated to fix missing commas in JSON output and missing $ in call to write_status info $EXIT_SUCCESS or $EXIT_UPDATE_REQUIRED

rpi-eeprom-update
```diff
@@ -320,25 +440,26 @@ removePreviousUpdates()
...
-   getUpdateVersion
-   if [ "${UPDATE_VERSION}" -gt "${CURRENT_VERSION}" ]; then
+   lookupVersionInfo
+
+   if [ "${BOOTLOADER_UPDATE_VERSION}" -gt "${BOOTLOADER_CURRENT_VERSION}" ]; then
       echo "*** UPDATE REQUIRED ***"
-      printVersions "${CURRENT_VERSION}" "${UPDATE_VERSION}"
-      write_status_info EXIT_UPDATE_REQUIRED "${CURRENT_VERSION}" "${UPDATE_VERSION}"
+      printVersions
+      write_status_info "${EXIT_UPDATE_REQUIRED}"
```
MilhouseVH: Did you really mean to change this to integers? I guess so, and now the comment on your previous commit makes sense, however the change `${1:-EXIT_FAILED}` does not as that will output a string if $1 is not defined. I can submit a PR if you let me know which you prefer - strings kinda made sense as if you decide to change the internal representation in future it doesn't affect the machine output (unless you change the strings, of course...)

timg236:
The shell exit code was designed to allow the caller to distinguish between no update required, update is required or an error. I think I originally intended to just map the data directly across to the machine output but missed the dollar so it appeared as a sting. Happy to accept a PR for whatever works best for LibreElec and make that the supported interface because AFAIK the machine interface not used by Raspbian yet.

MilhouseVH:
I'll push a PR which outputs the string `EXIT_SUCCESS` or `EXIT_UPDATE_REQUIRED` as the `EXITCODE` (no machine status file is created for any other type of exit code, ie fail) as that is what I've been expecting until now, thanks! If anyone wants to debate it in future I'm happy to have that conversation and can change to integers if that is considered a better solution.

PR #36, opened 2019-10-22
Restore machine output exit code as string

MilhouseVH:
Also ensure 0 as default bootloader timestamp for both JSON and non-JSON.


MilhouseVH:
Hang on... quick fix inbound.

timg236:
No problem. I’m away from a computer right now but will merge it tomorrow. I’m planning to bump the vl805 binary (but not the bootloader ) from beta to critical to start the vli update for everyone.
Once we are sure it’s safe we’ll add rpi-eeprom to the apt update packages so everyone gets this.

MilhouseVH:
OK thanks. I've applied the fix to this PR (and tested it) and it's working to my satisfaction now.

timg236:
Thanks looks good

PR #37, opened 2019-10-23
rpi-eeprom-update: Don't remove update files if run with no arguments

timg236:
Running 'sudo rpi-eeprom-update' will remove update files from the
boot partition. Since the user may run this command to check if
an update is required it will have the surprising side effect of
removing the pending update applied by the systemd service which
runs 'rpi-eeprom-update -a' at startup.

Move the cleanup code so that it only runs if a manual or automatic
update was requested.

Add an explicit option (-r) to just remove the previous update files.

### DIFF
--- a/rpi-eeprom-update
+++ b/rpi-eeprom-update
@@ -285,6 +285,7 @@ rpi-eeprom-update [options]... [FILE]
    -h Display help text and exit
    -j Write status information using JSON notation
    -m Write status information to the given file when run without -a or -f
+   -r Removes temporary EEPROM update files from the boot partition.
    -u Install the specified VL805 (USB EEPROM) image file.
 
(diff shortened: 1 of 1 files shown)


timg236:
@XECDesign @MilhouseVH 

This resolves a confusing issue where running 'sudo rpi-eeprom-update' to check the version removes the automatic update from Raspbian. 
I think LibreElec doesn't run 'rpi-eeprom-update -a' so it might be worth running with the '-r' option at startup to get the automatic cleanup.
Sorry for the interface change !

MilhouseVH:
We already clean-up on boot (manually, `rm -f /flash/pieeprom.{bin,upd,sig}` etc. in `init`) so this change shouldn't affect LibreELEC, but thanks for the heads up. Code changes look fine to me.

PR #38, opened 2019-10-23
critical update: Update VLI to 000137ab revision

timg236:
recovery.bin is also updated from the beta revision in order to apply
the VL805 EEPROM image.


timg236:
@XECDesign - please could you update the apt package update for this ?

PR #39, opened 2019-10-23
rpi-eeprom-update: Verify DPKG checksums if available

timg236:
If the rpi-eeprom-images checksums file is available then validate
the checksums before applying the update.

Add a -i option to ignore checksums, this should only be required for
developers or people who have cloned the rpi-eeprom Github and installed
new files which are not available in the apt package.

### DIFF
--- a/rpi-eeprom-update
+++ b/rpi-eeprom-update
@@ -139,7 +139,19 @@ applyRecoveryUpdate()
 }
 
 applyUpdate() {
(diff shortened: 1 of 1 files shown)


timg236:
@XECDesign added md5sums feature to guard against root-fs corruption

@MilhouseVH This shouldn't do anything on LibreElec because the dpkg files won't be there but please shout if you see a problem.

timg236:
Merging now so that we can build a test .deb file

XECDesign:
on internal

PR #44, opened 2019-10-28
Debian/buster

XECDesign:


timg236:
I tried the updated package on the internal repo and worked ok. Feel free to push to public when you are ready.

PR #57, opened 2019-11-21
rpi-eeprom-config - fix up help text (and therefore also man page)

andrum99:
The current help text for rpi-eeprom-config is unclear. This PR changes them so that the help text matches what the tool does and what each parameter specifies. The man page, automatically derived from the help text, is also fixed.

### DIFF
--- a/rpi-eeprom-config
+++ b/rpi-eeprom-config
@@ -113,7 +113,7 @@ combination of the original EEPROM image file, along with the new bootloader \
 configuration.')
     parser.add_argument('--config', help='Filename of new bootloader config')
     parser.add_argument('--out', help='Name of output file')
-    parser.add_argument('EEPROM', help='EEPROM filename (pieeprom.bin)')
+    parser.add_argument('eeprom', help='EEPROM filename (pieeprom.bin)')
     args = parser.parse_args()
 
     image = BootloaderImage(args.eeprom, args.out)
--- a/rpi-eeprom-config
+++ b/rpi-eeprom-config
@@ -98,9 +98,9 @@ class BootloaderImage(object):
 
 def main():
     parser = argparse.ArgumentParser('RPI EEPROM config tool')
-    parser.add_argument('--config', help='Filename of new bootloader config')
-    parser.add_argument('--out', help='Filename for the EEPROM image with updated config')
-    parser.add_argument('eeprom', help='EEPROM filename (pieeprom.bin)')
+    parser.add_argument('--config', help='Insert new bootloader config into EEPROM image. CONFIG specifies the file containing the new config.')
+    parser.add_argument('--out', help='Extract current bootloader config from EEPROM image. OUT specifies the file to put the config into.')
+    parser.add_argument('eeprom', help='Name of EEPROM image file (e.g. pieeprom.bin)')
     args = parser.parse_args()
 
     image = BootloaderImage(args.eeprom, args.out)
--- a/rpi-eeprom-config
+++ b/rpi-eeprom-config
@@ -97,10 +97,23 @@ class BootloaderImage(object):
                 sys.stdout.write(config_bytes)
 
 def main():
-    parser = argparse.ArgumentParser('RPI EEPROM config tool')
-    parser.add_argument('--config', help='Insert new bootloader config into EEPROM image. CONFIG specifies the file containing the new config.')
-    parser.add_argument('--out', help='Extract current bootloader config from EEPROM image. OUT specifies the file to put the config into.')
-    parser.add_argument('eeprom', help='Name of EEPROM image file (e.g. pieeprom.bin)')
+    parser = argparse.ArgumentParser(formatter_class=argparse.RawDescriptionHelpFormatter, \
+    description='RPI EEPROM config tool. \
+\n\nThere are 3 operating modes: \
+\n\n1. Output the bootloader configuration stored in an EEPROM image file to \
+the screen (STDOUT): specify only the name of an EEPROM image file using the \
+EEPROM option. \
+\n\n2. Output the bootloader configuration stored in an EEPROM image file to a \
+file: specify the EEPROM image file using the EEPROM option, and the output file\
+using the --out option. \
+\n\n3. Insert a new bootloader configuration into an EEPROM image file: specify \
+the EEPROM image file using the EEPROM option, the bootloader configuration file\
+using the --config option, and the name to use for the new file which will be a \
+combination of the original EEPROM image file, along with the new bootloader \
+configuration.')
+    parser.add_argument('--config', help='Filename of new bootloader config')
+    parser.add_argument('--out', help='Name of output file')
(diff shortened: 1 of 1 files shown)


rpi-eeprom-config
```diff
@@ -98,9 +98,9 @@ class BootloaderImage(object):
 
 def main():
     parser = argparse.ArgumentParser('RPI EEPROM config tool')
-    parser.add_argument('--config', help='Filename of new bootloader config')
-    parser.add_argument('--out', help='Filename for the EEPROM image with updated config')
-    parser.add_argument('eeprom', help='EEPROM filename (pieeprom.bin)')
+    parser.add_argument('--config', help='Insert new bootloader config into EEPROM image. CONFIG specifies the file containing the new config.')
+    parser.add_argument('--out', help='Extract current bootloader config from EEPROM image. OUT specifies the file to put the config into.')
```
timg236: Changes are an improvement but I've spotted that '--out' is not quite correct. It's just the name of the output file which can be the EEPROM if run as 
rpi-eeprom-config --config new-config.txt --out new-eeprom.bin current-eeprom.bin

andrum99: Ah, OK. So --out is used for two options - when run with just --out as a switch, it dumps the config, and when run in combination with --config it instead copies an existing EEPROM image, rewriting the config as specified by CONFIG in the process - is that correct?

timg236: Yes, '--out' specifies a filename for the output, if --out is omitted then the output is written to stdout so

These commands are equivalent
rpi-eeprom-config pieeprom.bin > boot.conf 
rpi-eeprom-config pieeprom.bin --out boot.conf 

and so are these
rpi-eeprom-config pieeprom.bin --config boot.conf > new-eeprom.bin
rpi-eeprom-config pieeprom.bin --config boot.conf --out new-eeprom.bin

Perhaps it would have been better if I hadn't added --out and called --config something like --update-config. Oh well :)

andrum99: It makes sense to have an option to output to a file, since noobs don't know you can pipe output to a file.

I will amend my PR.

andrum99:
I'm missing the bit that turns the relevant parts of the script into the man page. I need to see that, since in order to correctly document the function of the script I need to insert more text into the 'description' part of the man page.

timg236:
It’s in the Debian/buster branch of this repo

andrum99:
Found it- help2man is the usage to man page tool.

andrum99:
OK that should be it sorted now. The indentation of the description text looks wrong, but it needs to be that way otherwise the indents within the source on those lines gets embedded in the help text, since I'm using an alternate text formatter in order to be able to insert my own newlines in the description to make it much more readable.

andrum99:
Forgot to mention - this PR also fixes a minor bug in the code, where the creation of the 'parser' object was done slightly wrong - the 'description=' bit was missing, which was causing the usage to include the text that was intended as part of the description to be output as if it was the name of the script, if you see what I mean.

rpi-eeprom-config
```diff
@@ -97,10 +97,23 @@ class BootloaderImage(object):
...
+\n\n3. Insert a new bootloader configuration into an EEPROM image file: specify \
+the EEPROM image file using the EEPROM option, the bootloader configuration file\
+using the --config option, and the name to use for the new file which will be a \
+combination of the original EEPROM image file, along with the new bootloader \
+configuration.')
     parser.add_argument('--config', help='Filename of new bootloader config')
-    parser.add_argument('--out', help='Filename for the EEPROM image with updated config')
-    parser.add_argument('eeprom', help='EEPROM filename (pieeprom.bin)')
+    parser.add_argument('--out', help='Name of output file')
+    parser.add_argument('EEPROM', help='EEPROM filename (pieeprom.bin)')
```
timg236: I think this should be '--eeprom'. Currently the unit test fails

cd  test
./test-rpi-eeprom-config

check_update ../firmware/critical/pieeprom-2019-07-15.bin bootconf-2019-07-15.txt
Traceback (most recent call last):
  File "/home/tim/rpi-eeprom/test/../rpi-eeprom-config", line 126, in <module>
    main()
  File "/home/tim/rpi-eeprom/test/../rpi-eeprom-config", line 119, in main
    image = BootloaderImage(args.eeprom, args.out)
AttributeError: 'Namespace' object has no attribute 'eeprom'
ERROR
Config-read: checksum mismatch

andrum99: Sorry that was me - I was toying with the idea of putting the EEPROM parameter in upper case in order to make the help description slightly more readable, but I decided against it, then forgot I had already changed that. It's back the way it was now - 'eeprom' in lower case, which is correct. (There is no --eeprom parameter).

timg236:
Thanks fix looks good

rpi-eeprom-config
```diff
@@ -97,9 +97,22 @@ class BootloaderImage(object):
...
+\n\nThere are 3 operating modes: \
+\n\n1. Output the bootloader configuration stored in an EEPROM image file to \
+the screen (STDOUT): specify only the name of an EEPROM image file using the \
+EEPROM option. \
+\n\n2. Output the bootloader configuration stored in an EEPROM image file to a \
+file: specify the EEPROM image file using the EEPROM option, and the output file\
+using the --out option. \
+\n\n3. Insert a new bootloader configuration into an EEPROM image file: specify \
+the EEPROM image file using the EEPROM option, the bootloader configuration file\
+using the --config option, and the name to use for the new file which will be a \
```
lurch: Should this have mentioned something like "and the name of the new file using the --out option." ?

Also, only a minor nitpick but I wonder if the text as-is might be misread as "a combination of the original EEPROM image file, along with the new bootloader configuration" being used to automatically generate "the name to use for the new file" ? :man_shrugging: (IYSWIM)

lurch:
Hello, nice to see you again @andrum99 ! :wave: 
Do you mind if I ask why you keep deleting and then recreating your GitHub account? :confused: 
It resulted in a bunch of "orphaned" PRs in the documentation repo, which then became slightly trickier to manage...

Thanks again for all your contributions though :+1: 

cc @JamesH65

rpi-eeprom-config
andrum99: Yes you're quite right. Haven't checked your second point but I'm sure you're right on that too.

andrum99:
> Hello, nice to see you again @andrum99 ! 
> Do you mind if I ask why you keep deleting and then recreating your GitHub account? 
> It resulted in a bunch of "orphaned" PRs in the documentation repo, which then became slightly trickier to manage...
> 
> Thanks again for all your contributions though 
> 
> cc @JamesH65

I've tended to want to cool off for a period occasionally, but I'm aware it messes things up when I've got active PR's open. It's not something I plan to repeat.

rpi-eeprom-config
andrum99: I will raise an additional PR to address your comments.

PR #59, opened 2019-11-22
Fix up help text per @lurch's comments in #57

andrum99:
Operating mode 3 failed to mention that the output file should be specified using the --out parameter.

### DIFF
--- a/rpi-eeprom-config
+++ b/rpi-eeprom-config
@@ -117,7 +117,7 @@ source EEPROM image.\
  which installs them to the /lib/firmware/raspberrypi/bootloader directory.')
     parser.add_argument('--config', help='Name of bootloader configuration file')
     parser.add_argument('--out', help='Name of output file')
-    parser.add_argument('eeprom', help='name of EEPROM file to use as input')
+    parser.add_argument('eeprom', help='Name of EEPROM file to use as input')
     args = parser.parse_args()
 
     image = BootloaderImage(args.eeprom, args.out)
--- a/rpi-eeprom-config
+++ b/rpi-eeprom-config
@@ -98,22 +98,26 @@ class BootloaderImage(object):
 
 def main():
     parser = argparse.ArgumentParser(formatter_class=argparse.RawDescriptionHelpFormatter, \
-    description='RPI EEPROM config tool. \
+    description='Bootloader EEPROM configuration tool for the Raspberry Pi 4B. \
 \n\nThere are 3 operating modes: \
 \n\n1. Output the bootloader configuration stored in an EEPROM image file to \
 the screen (STDOUT): specify only the name of an EEPROM image file using the \
-EEPROM option. \
+\'eeprom\' option. \
 \n\n2. Output the bootloader configuration stored in an EEPROM image file to a \
-file: specify the EEPROM image file using the EEPROM option, and the output file\
-using the --out option. \
+file: specify the EEPROM image file using the \'eeprom\' option, and the output \
+file using the \'--out\' option.\
 \n\n3. Insert a new bootloader configuration into an EEPROM image file: specify \
-the EEPROM image file using the EEPROM option, the bootloader configuration file\
-using the --config option. A new file which is a combination of the EEPROM image\
-file, together with the new bootloader configuration file you specify will be\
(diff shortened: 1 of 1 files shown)


andrum99:
OK this is ready for review now. Some other minor changes:

- Fix up single sentence description of the tool
- Minor fix up to operating mode 2
- In description, put parameters within single quotes. The alternative would have been to change the EEPROM option to capital letters, since the option must be highlighted somehow within the description text, either as caps or within quotes.

rpi-eeprom-config
```diff
@@ -98,22 +98,26 @@ class BootloaderImage(object):
...
-using the --config option, and the name to use for the new file which will be a \
-combination of the original EEPROM image file, along with the new bootloader \
-configuration.')
+the source EEPROM image file using the \'eeprom\' option and the bootloader \
+configuration file using the \'--config\' option. A new file which is a \
+combination of the EEPROM image file, together with the new bootloader \
+configuration file will be created - specify its name using the \'--out\' option. \
+The new bootloader configuration will replace any configuration present in the \
+source EEPROM image.\
+\n\nBootloader EEPROM images are contained in the \'rpi-eeprom-images\' package,\
```
lurch: Would "are supplied by the 'rpi-eeprom-images' package" sound better than "are contained in the 'rpi-eeprom-images' package" ? :man_shrugging:

rpi-eeprom-config
```diff
@@ -98,22 +98,26 @@ class BootloaderImage(object):
...
+ which installs them to the /lib/firmware/raspberrypi/bootloader directory.')
     parser.add_argument('--config', help='Filename of new bootloader config')
     parser.add_argument('--out', help='Name of output file')
-    parser.add_argument('eeprom', help='EEPROM filename (pieeprom.bin)')
+    parser.add_argument('eeprom', help='name of EEPROM file to use as input')
```
lurch: It looks like all three command-line arguments expect a filename? Maybe it'd be more consistent to use either "name" or "filename" in the help-text for all three?

rpi-eeprom-config
andrum99: Yes, although it appears the text I wrote already says that. (Unless you've somehow updated it, and pushed it to my internal git repo 😛)

rpi-eeprom-config
andrum99: Yes, and I've also removed the word "new" to simplify it, since the explanation of the parameters is not enough on its own to understand how the tool actually works, due to how the parameters operate. The description of the parameters must be read together with the description, which describes the 3 operating modes.

rpi-eeprom-config
```diff
@@ -98,22 +98,26 @@ class BootloaderImage(object):
...
-the EEPROM image file using the EEPROM option, the bootloader configuration file\
-using the --config option, and the name to use for the new file which will be a \
-combination of the original EEPROM image file, along with the new bootloader \
-configuration.')
+the source EEPROM image file using the \'eeprom\' option and the bootloader \
+configuration file using the \'--config\' option. A new file which is a \
+combination of the EEPROM image file, together with the new bootloader \
+configuration file will be created - specify its name using the \'--out\' option. \
+The new bootloader configuration will replace any configuration present in the \
+source EEPROM image.\
```
lurch: This feels incredibly wordy for the `--help` output of a tool :confused: (or maybe that's just me?!)

If the "Insert a new bootloader configuration..." line was changed to "Inject a new bootloader configuration..." do you think it would still make sense if the "A new file which is a combination of the EEPROM image file, together with the new bootloader configuration file will be created" part was removed? :man_shrugging:

andrum99: It is wordy, but that was mainly because I was trying to avoid having to write a separate man page. You're also not really injecting something, since the new configuration is written to a new file. I'm happy with it as-is, on this one. I will leave it to the repo owner to decide which approach they want to go for.

rpi-eeprom-config
lurch: Should the 'name of EEPROM file' start with a capital N ?

timg236:
Looks good to me. My general preference is to keep the documentation in the code which is why these changes are much appreciated :)

PR #61, opened 2019-11-29
vcgencmd: Use cpuinfo instead of otp_dump and add check for vcgencmd …

timg236:


timg236:
@XECDesign

rpi-eeprom-update
```diff
@@ -224,16 +224,27 @@ getBootloaderUpdateVersion() {
...
    fi
 
-   CPU_VER="$(vcgencmd otp_dump | grep 30: | cut -c8)"
-
-   if [ "${CPU_VER}" != "3" ]; then
-      # Not a BCM2711, no EEPROMs to update.
-      exit ${EXIT_SUCCESS}
-   fi
+   # Exit if VCHI is stuck to avoid blocking apt updates
+   nohup vcgencmd version > /dev/null 2>&1 &
```
XECDesign: In the failing case, vcgencmd would return with 'VCHI initialization failed'. I'm not sure that this test does anything.

timg236: ok, I'll split these changes in-case the otp_dump change is sufficient

PR #65, opened 2019-12-09
Add automated custom EEPROM configuration

m-roberts:
### Functionality/purpose
This PR extends the functionality of `rpi-eeprom-update` to support setting custom EEPROM configuration values at the point of updating firmware. This allows users to ensure that any future changes to default bootloader configuration values (as happened with `WAKE_ON_GPIO`) that need to stay the same for their particular setup do not cause any issues.

The need for this functionality arose from the change to the WAKE_ON_GPIO flag within the RPi EEPROM for improved HAT support. This prevents the pi-top [4] from being able to shut down as it uses the voltage drop on the Pi to know that the shutdown sequence had been completed.

This PR provides users of pi-top [4] to have an easy way to ensure that the `WAKE_ON_GPIO` and `POWER_OFF_ON_HALT` values on the EEPROM (and potentially others in the future) are set correctly directly from the OS. It also allows any user to set any bootloader configuration variable from the OS image.

### Methodology
This PR achieves this functionality by reading the custom EEPROM config from `/etc/default/rpi-eeprom-config`, if no file is specified in `/etc/default/rpi-eeprom-update` (via `CUSTOM_EEPROM_CONFIG`), if the file exists. This file is intended to be formatted identically to the output of `vcgencmd bootloader_config`. These values are applied to the EEPROM config to be programmed, if the user has not asked to use the defaults (using the `-d` flag).

### How this is done in the code
In the current codebase (before this PR), [fewer than 3 lines of config](https://github.com/raspberrypi/rpi-eeprom/blob/master/rpi-eeprom-update#L88) (capturing empty configs and early versions which do not have the new default value for `WAKE_ON_GPIO`) [sets `OVERWRITE_CONFIG=1`](https://github.com/raspberrypi/rpi-eeprom/blob/master/rpi-eeprom-update#L92). This would then [skip](https://github.com/raspberrypi/rpi-eeprom/blob/master/rpi-eeprom-update#L97) injecting [current config values into the EEPROM file to be installed](https://github.com/raspberrypi/rpi-eeprom/blob/master/rpi-eeprom-update#L98).

I felt that it was better, in light of the changes proposed here, to [take the 'default' EEPROM values from the firmware to be installed](https://github.com/raspberrypi/rpi-eeprom/pull/65/files#diff-86eaef688a7bf372514e072950593ed7R99), and - if the user has not opted to simply use the defaults - then [patch the values from the current config](https://github.com/raspberrypi/rpi-eeprom/pull/65/files#diff-86eaef688a7bf372514e072950593ed7R111-R115), followed by the [overrides specified in `CUSTOM_EEPROM_CONFIG`](https://github.com/raspberrypi/rpi-eeprom/pull/65/files#diff-86eaef688a7bf372514e072950593ed7R117-R137).

This felt like a cleaner approach for modifying the config to apply to the firmware file, as it more clearly captures what is happening for each use case, and simplifies extensibility.

After all of these changes have (or indeed have not) been applied, [the EEPROM configuration is then printed out to screen](https://github.com/raspberrypi/rpi-eeprom/pull/65/files#diff-86eaef688a7bf372514e072950593ed7R139-R141), to clearly show what is to be applied in the update.

One important 'caveat' is that this code will [ignore `FREEZE_VERSION` from the config file](https://github.com/raspberrypi/rpi-eeprom/pull/65/files#diff-86eaef688a7bf372514e072950593ed7R126-R128), as it doesn't make sense to install an updated firmware which _then_ locks its version. If the user does not want updates, then they should disable the auto updater systemd service.

If a field specified in the custom EEPROM config is not detected, [a warning will be printed](https://github.com/raspberrypi/rpi-eeprom/pull/65/files#diff-86eaef688a7bf372514e072950593ed7R133-R134) to notify the user that this will not be applied to the config.

#### Help text
* [Added instructions](https://github.com/raspberrypi/rpi-eeprom/pull/65/files#diff-86eaef688a7bf372514e072950593ed7R377-R382) for how to use this mechanism to help text.
* Clarified wording: ["The syntax is the same as config.txt"](https://github.com/raspberrypi/rpi-eeprom/pull/65/files#diff-86eaef688a7bf372514e072950593ed7L331-L332) - now specifically mentions that this [refers to bootloader configuration](https://github.com/raspberrypi/rpi-eeprom/pull/65/files#diff-86eaef688a7bf372514e072950593ed7R384-R385) syntax.
* Clarified wording: when `-d` is used, the "default" bootloader config that is referred to is [what is provided in the EEPROM image to be installed](https://github.com/raspberrypi/rpi-eeprom/pull/65/files#diff-86eaef688a7bf372514e072950593ed7R354).
#### Config file format checking
* ~~Use [`shellcheck`](https://github.com/raspberrypi/rpi-eeprom/pull/65/files#diff-86eaef688a7bf372514e072950593ed7R119) to check formatting of config files in `/etc/default/`~~
**~~NOTE: this adds a dependency, which may not be desirable~~**
#### Variable naming
* Rename `OVERWRITE_CONFIG` to `USE_EEPROM_INTERNAL_CONFIG`. I felt that this variable was confusingly named, so I renamed it to better capture what is happening (in my opinion).

### Other changes not directly related to the detailed functionality
(NOTE: these may be better suited for a separate PR)

* [Remove argument](https://github.com/raspberrypi/rpi-eeprom/pull/65/files#diff-86eaef688a7bf372514e072950593ed7L173-R218) from call to `prepareImage` in `applyUpdate`
* [Fix typo](https://github.com/raspberrypi/rpi-eeprom/pull/65/files#diff-86eaef688a7bf372514e072950593ed7L175-R220) in applyUpdate: `bootloaer` --> `bootloader`
* [Add double quotes](https://github.com/raspberrypi/rpi-eeprom/pull/65/files#diff-86eaef688a7bf372514e072950593ed7L599-R652) in `grep "FREEZE_VERSION=1"`
* [Fix indentation](https://github.com/raspberrypi/rpi-eeprom/pull/65/files#diff-86eaef688a7bf372514e072950593ed7L116-R174) in `applyRecoveryUpdate`
* [Remove unnecessary](https://github.com/raspberrypi/rpi-eeprom/pull/65/files#diff-86eaef688a7bf372514e072950593ed7L118) `TMP_EEPROM_IMAGE="$(mktemp)"` in `applyRecoveryUpdate` (at least, I think that this is unnecessary?)

### Possible future extensions
This could be extended to check the current bootloader configuration against the desired custom bootloader configuration, and notify the user that they may get unexpected behaviour without applying the changes to the RPi firmware. This could be useful in mass deployments where many Raspberry Pi units are used against one OS image, such as over network boot, for example.

### DIFF
--- a/rpi-eeprom-update
+++ b/rpi-eeprom-update
@@ -22,6 +22,8 @@ RECOVERY_BIN=${RECOVERY_BIN:-${FIRMWARE_ROOT}/${FIRMWARE_RELEASE_STATUS}/recover
 BOOTFS=${BOOTFS:-/boot}
 VCMAILBOX=${VCMAILBOX:-/opt/vc/bin/vcmailbox}
 
+CUSTOM_EEPROM_CONFIG=${CUSTOM_EEPROM_CONFIG:-/etc/default/rpi-eeprom-config}
+
 EXIT_SUCCESS=0
 EXIT_UPDATE_REQUIRED=1
 EXIT_FAILED=2
@@ -29,7 +31,7 @@ EXIT_EEPROM_FROZEN=3
 # Reserved
 # EXIT_PREVIOUS_UPDATE_FAILED=4
 
-OVERWRITE_CONFIG=0
+USE_EEPROM_INTERNAL_CONFIG=0
 # Maximum safe SPI speed for EEPROM access 16000, slower is ok.
 SPI_SPEED=16000
 # Timestamp for first release which doesn't have a timestamp field
@@ -54,8 +56,11 @@ cleanup() {
    if [ -f "${TMP_EEPROM_IMAGE}" ]; then
       rm -f "${TMP_EEPROM_IMAGE}"
    fi
-   if [ -f "${TMP_EEPROM_CONFIG}" ]; then
-      rm -f "${TMP_EEPROM_CONFIG}"
+   if [ -f "${TMP_CURRENT_EEPROM_CONFIG}" ]; then
+      rm -f "${TMP_CURRENT_EEPROM_CONFIG}"
+   fi
+   if [ -f "${TMP_NEW_EEPROM_CONFIG}" ]; then
+      rm -f "${TMP_NEW_EEPROM_CONFIG}"
    fi
    if [ -d "${TMP_BOOTFS_MNT}" ]; then
       umount "${TMP_BOOTFS_MNT}"
@@ -63,7 +68,8 @@ cleanup() {
    fi
    TMP_BOOTFS_MNT=
    TMP_EEPROM_IMAGE=
-   TMP_EEPROM_CONFIG=
+   TMP_CURRENT_EEPROM_CONFIG=
+   TMP_NEW_EEPROM_CONFIG=
 }
 trap cleanup EXIT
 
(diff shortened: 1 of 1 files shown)


XECDesign:
Not a fan of adding a 20MB dependency (shellcheck) to a tool that is shipped in Raspbian lite.

MilhouseVH:
`shellcheck` isn't/won't be supported by LibreELEC.

m-roberts:
I will remove then! Just wanted to throw it out there as I thought it might have some use. Please let me know how you feel about the rest of the commit!

rpi-eeprom-update
```diff
@@ -54,50 +61,85 @@ cleanup() {
...
+      rpi-eeprom-config "${TMP_EEPROM_IMAGE}" --out "${TMP_NEW_EEPROM_CONFIG}"
+
+      # Clean up empty lines
+      sed -i -e 's/[\t ]//g;/^$/d' "${TMP_NEW_EEPROM_CONFIG}"
+
+      no_of_eeprom_config_properties="$(wc -l "${TMP_CURRENT_EEPROM_CONFIG}" | awk '{print $1}')"
+      if [ "$no_of_eeprom_config_properties" -lt 3 ]; then
+         # Prevent the initial bootloader config with WAKE_ON_GPIO=0
+         # propagating to newer versions by accident
+         sed -i 's/WAKE_ON_GPIO=.*/WAKE_ON_GPIO=1/1' "${TMP_NEW_EEPROM_CONFIG}"
```
```suggestion
         sed -i -e 's/WAKE_ON_GPIO=.*/WAKE_ON_GPIO=1/1' "${TMP_NEW_EEPROM_CONFIG}"
```

rpi-eeprom-update
```diff
@@ -54,50 +61,85 @@ cleanup() {
...
+      fi
+
+      # Apply current config settings to new EEPROM config
+      while read line; do
+         field=$(echo "${line}" | cut -d'=' -f1)
+         sed -i "s/${field}=.*/${line}/1" "${TMP_NEW_EEPROM_CONFIG}"
```
```suggestion
         sed -i -e "s/${field}=.*/${line}/1" "${TMP_NEW_EEPROM_CONFIG}"
```

rpi-eeprom-update
```diff
@@ -54,50 +61,85 @@ cleanup() {
...
+         echo "\n*** APPLYING CUSTOM EEPROM CONFIGURATION ***"
+         while read line; do
+            field="$(echo "${line}" | cut -d'=' -f1)"
+
+            if grep -q "${field}" "${TMP_NEW_EEPROM_CONFIG}"; then
+               if [ "${field}" = "FREEZE_VERSION" ]l; then
+                  echo "WARNING: Skipping ${line} from custom EEPROM configuration..."
+                  echo "If you wish to prevent firmware updates from the OS, disable the rpi-eeprom-update systemd service."
+               else
+                  sed -i "s/${field}=.*/${line}/1" "${TMP_NEW_EEPROM_CONFIG}"
```
```suggestion
                  sed -i -e "s/${field}=.*/${line}/1" "${TMP_NEW_EEPROM_CONFIG}"
```

andrum99:
Would it not be better to just have two options: 1. Use the config specified in the updated EEPROM image, and 2. Copy the bootloader EEPROM config over from the current image to the new one? And if 2 is selected, write what the new config would have been to a file and print a message on screen. This has the advantage that it is similar to how Debian handles updated config files, and also keeps things fairly simply. This PR seems a bit complex.

rpi-eeprom-update
```diff
@@ -54,50 +61,85 @@ cleanup() {
...
+      done < "${TMP_CURRENT_EEPROM_CONFIG}"
+
+      # Apply user customisations to new EEPROM config
+      if [ -f "${CUSTOM_EEPROM_CONFIG}" ]; then
+         echo "\n*** APPLYING CUSTOM EEPROM CONFIGURATION ***"
+         while read line; do
+            field="$(echo "${line}" | cut -d'=' -f1)"
+
+            if grep -q "${field}" "${TMP_NEW_EEPROM_CONFIG}"; then
+               if [ "${field}" = "FREEZE_VERSION" ]l; then
```
lurch: Is that `l` just before the semi-colon a typo?

timg236:
I've had a bit more time to read the change now, I think what it's trying to do is reasonable shouldn't be in the core script. Most users won't need this functionality but for those who do there will be never ending feature creep.

Instead, I'll push a PR shortly which adds a hook (think Git pre-commit hook). If the hook is present rpi-eeprom-update will pipe the config through this script and it can change it however it like. 
The hook will also receive the name of the source binary as context data.

Always happy to see improvements to help text etc but these must be in separate PRs for review-ability.

lurch:
> pipe the config through this script and it can change it however it like.

Sounds like a good idea. It seems the change in this PR is effectively merging two dictionaries, and to me it feels like that'd be something better done in a "proper" scripting language, rather than trying to hack it into bash using grep and sed.

lurch:
@m-roberts Does #66 allow you to do what you want?

rpi-eeprom-update
m-roberts: Yes 😓

m-roberts:
At a cursory glance, yes. I will play around with this shortly to confirm!

m-roberts:
I can confirm that this works for me.

timg236:
Thanks @m-roberts 

@XECDesign if you are around tomorrow please could you create a .deb update on internal ?

PR #66, opened 2019-12-10
Eeprom config hook

timg236:
Add a new hook script and basic documentation for this.

Example
Add EEPROM_CONFIG_HOOK=/usr/local/bin/my-pi-eeprom-hook to /etc/default/rpi-eeprom-update

cat /usr/local/bin/my-eeprom-config-hook
#!/bin/sh
while read line;
do
  echo ${line} | sed 's/WAKE_ON_GPIO.*/WAKE_ON_GPIO=0/g'
done


rpi-eeprom-update
```diff
@@ -85,7 +90,16 @@ prepareImage()
    backup="${FIRMWARE_BACKUP_DIR}/pieeprom-backup-$(date +%Y%m%d-%H%M%S).conf"
    cp -f "${TMP_EEPROM_CONFIG}" "${backup}"
 
-   if [ "$(wc -l "${TMP_EEPROM_CONFIG}" | awk '{print $1}')" -lt 3 ]; then
+   if [ -x "${EEPROM_CONFIG_HOOK}" ]; then
+      echo "Running EEPROM config hook ${EEPROM_CONFIG_HOOK}"
+      if ! "${EEPROM_CONFIG_HOOK}" -u "${BOOTLOADER_UPDATE_IMAGE}" < "${TMP_EEPROM_CONFIG}" > "${NEW_EEPROM_CONFIG}"; then
+         echo "EEPROM config hook \"${EEPROM_CONFIG_HOOK}\"  failed. Using original configuration"
```
lurch: Does "Using original configuration" imply that another `cp -f "${TMP_EEPROM_CONFIG}" "${NEW_EEPROM_CONFIG}"` is needed here?

timg236: good point, must add all the lines to Git before pushing :)

rpi-eeprom-update
```diff
@@ -319,20 +330,42 @@ rpi-eeprom-update [options]... [FILE]
...
+output must contain at least 3 lines and should contain WAKE_ON_GPIO
+and POWER_OFF_ON_HALT settings.
+
+USE_FLASHROM
+
+The flashrom update mechanism may be enabled by setting USE_FLASHROM=1. This
+also selects the vl805 tool instead of using recovery.bin to perform the
+update. This may be desirable if an immediate update is required or if an
+sd-card is not present.
+Howver, this not recommended because the SPI pins are muxed with audio and other
```
andrum99: Howver -> However

m-roberts:
It might be nice to have a way to force installation of the latest bootloader (when coupled with `-a`), in case its config does not match that which would be applied by the hook. Obviously `-f` is taken, so I am not sure what would be a good flag. But it would be nice to trigger the application of the hook, in case of migrating the SD card image to a Raspberry Pi which has updated to the latest firmware, but does not have the desired configuration.

timg236:
Merging this now. I think the forcing of the latest bootloader + customisation should be probably be a higher level operation e.g. an extra systemd service which depends on rpi-eeprom but is able to inspect the bootloader config and force an update if necessary. Users can then install this package to apply customizations or remove it for vanilla Raspbian behaviour.

PR #67, opened 2019-12-12
rpi-eeprom-update: Improve version selection for VL805

timg236:
Deprecate the vl805.latest file and compare version numbers. This
ensures that the rpi-eeprom-update will never downgrade the VL805.

### DIFF
--- a/rpi-eeprom-update
+++ b/rpi-eeprom-update
@@ -435,19 +435,17 @@ getVL805CurrentVersion()
 
 getVL805UpdateVersion()
 {
-   # The latest VL805 version is indicated by a file containing the version
-   # number. If the version file exists then verify that the corresponding
-   # VL805 binary exists before selecting it.
-   # There are no user modifiable sections in the VL805 firmware and there 
-   # are unlikely to be many updates to it so just indicate the latest supported
-   # version. This also avoids making any assumptions about the VLI version numbers.
+   # The VL805 version number is an eight character hex string. Select the
+   # largest number for the newest version.
+   # The vl805.latest version is retained for backwards compatibility with
+   # thirdparty scripts (are there any?) but it not used by this script and
+   # is deprecated.
    VL805_UPDATE_VERSION=""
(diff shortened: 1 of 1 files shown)


timg236:
@MilhouseVH @XECDesign Not particularly urgent but doing this in preparation for possible VL805 future firmware upgrades to ensure that rpi-eeprom-update won't automatically downgrade the firmware. The hope that the 137ab would be the last version was a bit optimistic :)

MilhouseVH:
Thanks @timg236 - we'll stop installing vl805.latest once this change lands so if you decide to remove it in future it won't be an issue for LibreELEC.

rpi-eeprom-update
```diff
@@ -468,7 +466,7 @@ lookupVersionInfo()
    fi
 
    if [ -n "${VL805_CURRENT_VERSION}" ] && [ -n "${VL805_UPDATE_VERSION}" ]; then
-      if [ "${VL805_CURRENT_VERSION}" != "${VL805_UPDATE_VERSION}" ]; then
+      if [ "${VL805_CURRENT_VERSION}" \< "${VL805_UPDATE_VERSION}" ];  then
```
lurch: It's been a while since I've done any bash scripting, but is the backslash needed here? https://linuxize.com/post/how-to-compare-strings-in-bash/#lexicographic-comparison seems to suggest it isn't.
(and I guess there's only one space needed before the `then` :wink: )

MilhouseVH: The article in your link is using `[[ ]]` notation for the examples and is discussing `bash` not `sh` - the latter (which is used by this script) doesn't support `[[ ]]` as these are non-POSIX `bash`-extensions. With single `[ ]` the `<` needs to be escaped to prevent it being interpreted as input redirection.

lurch: Oh, how embarrassing - I hadn't spotted that this was using `sh` rather than `bash` :man_facepalming:

PR #76, opened 2020-01-10
rpi-eeprom-update: Add support for board-revision specific images.

timg236:
Add the option for board-revision specific bootloaders firmware. This
allows critical firmware updates to be applied to specific board
revisions first before these are promoted to the generic firmware
directories.

Update the find operation to follow symlinks so that EEPROM images
in the board revision specific subdirectories can be links to the
generic images.


timg236:
@XECDesign @MilhouseVH 
I think this should be mostly harmless and it isn't required right not but it's good to get the update infrastructure tested before we need to update EEPROMs for real :)

PR #77, opened 2020-01-13
Debian/buster

XECDesign:


debian/changelog
```diff
@@ -1,3 +1,27 @@
+rpi-eeprom (3.0-1) buster; urgency=medium
+
+  [ Tim Gover ]
+  * vcgencmd: Use cpuinfo instead of otp_dump
+  * pieeprom-2019-12-03: Fix DHCP Option43 menu handling
+  * install: Fix path to vl805 executable
+  * rpi-eeprom-update: Add support for config txt processing hooks
+  * rpi-eeprom-update: Add hook/flashrom help and fix section formatting
```
m-roberts: You missed out the commit `rpi-eeprom-update: Add support for config txt processing hooks`, which is what this commit is actually referring to.

XECDesign: Seems to be in the commits tab of this PR as far as I can tell.

m-roberts: Fair enough

PR #79, opened 2020-01-14
rpi-eeprom-config: Fix padding calculation

timg236:


timg236:
@XECDesign doesn't fail with existing EEPROMs but I noticed a failure on a debug build. Will try to make sure new images don't trigger the bug but we ought to fix the script anyway.

PR #84, opened 2020-01-17
2019-01-17: Handle DHCP option 0 (padding)

timg236:


timg236:
Merging now for SD card voltage detect

PR #91, opened 2020-02-11
Clarifying the Dated .bin files

jpartain89:
Added extra clarification for the beta `.bin` file, removing the static date from the command, as plenty of people would definitely copy/paste that blindly without realizing there are multiple versions in that directory. (Me being one of those people)

### DIFF
--- a/firmware/raspberry_pi4_network_boot_beta.md
+++ b/firmware/raspberry_pi4_network_boot_beta.md
@@ -40,9 +40,11 @@ sudo echo FIRMWARE_RELEASE_STATUS="beta" > /etc/default/rpi-eeprom-update
 
 ### Configuration - Enable network boot
 Network boot is not enabled by default in the bootloader. To enable it the bootloader configuration file must be edited.
+
+**NOTE:** The `/lib/firmware...` line below ends with a date at the end of the file's name. We have a few versions of the beta bootloader in there, so make sure to select the most up-to-date version for your raspberry pi.
 ```
 # Extract the configuration file
-cp /lib/firmware/raspberrypi/bootloader/beta/pieeprom-2019-09-23.bin pieeprom.bin
+cp /lib/firmware/raspberrypi/bootloader/beta/pieeprom-$DATE.bin pieeprom.bin
 rpi-eeprom-config pieeprom.bin > bootconf.txt
 
 # Enable network boot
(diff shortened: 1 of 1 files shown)


firmware/raspberry_pi4_network_boot_beta.md
```diff
@@ -40,9 +40,11 @@ sudo echo FIRMWARE_RELEASE_STATUS="beta" > /etc/default/rpi-eeprom-update
 
 ### Configuration - Enable network boot
 Network boot is not enabled by default in the bootloader. To enable it the bootloader configuration file must be edited.
+
+**NOTE:** The `/lib/firmware...` line below ends with a date at the end of the file's name. We have a few versions of the beta bootloader in there, so make sure to select the most up-to-date version for your raspberry pi.
```
lurch: Personally, I'd change "We have a few ..." to "There are several ..."

And "Raspberry Pi" ought to be capitalised :wink:

jpartain89: Made those changes for you. Definitely makes more sense that way. Thanks!

firmware/raspberry_pi4_network_boot_beta.md
```diff
@@ -40,9 +40,11 @@ sudo echo FIRMWARE_RELEASE_STATUS="beta" > /etc/default/rpi-eeprom-update
 
 ### Configuration - Enable network boot
 Network boot is not enabled by default in the bootloader. To enable it the bootloader configuration file must be edited.
+
+**NOTE:** The `/lib/firmware...` line below ends with a date at the end of the file's name. There are several versions of the beta bootloader in there, so make sure to select the most up-to-date version for your Raspberry Pi.
 ```
 # Extract the configuration file
-cp /lib/firmware/raspberrypi/bootloader/beta/pieeprom-2019-09-23.bin pieeprom.bin
+cp /lib/firmware/raspberrypi/bootloader/beta/pieeprom-$DATE.bin pieeprom.bin
```
timg236: I think $DATE will not do the what the user expects if this is pasted because it will be expanded by the shell into either the wrong date or the empty string. Perhaps pieeprom-LATEST.bin would be obvious enough

jpartain89: This is that fun gray area of "whats the best way to convey this information to a new user".... 

I'm OK with whatever you think would be the best way to convey the intended information.

lurch: ...or `pieeprom-YYYY-MM-DD.bin` :man_shrugging:

jpartain89:
Changed the verbiage in the `note` block to be more deliberate and less clumsy in explaining the date formatting, along with changing to YYYY-MM-DD for the file name.

PR #96, opened 2020-02-27
rpi-eeprom-update: Remove dependency on vl805

timg236:
The vl805 utility is not available for aarch64 and is deprecated
because it's only required for USE_FLASHROM.

Remove the dependency check so that it can be removed from the package.

The USE_FLASHROM code will be removed at a later date once the
self-update mechanism in the bootloader EEPROM is released.

### DIFF
--- a/rpi-eeprom-update
+++ b/rpi-eeprom-update
@@ -262,10 +262,6 @@ checkDependencies() {
       FIRMWARE_IMAGE_DIR="${FIRMWARE_IMAGE_DIR}-${BOARD_REVISION}"
    fi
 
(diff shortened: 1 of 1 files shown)


rpi-eeprom-update
```diff
@@ -262,10 +262,6 @@ checkDependencies() {
       FIRMWARE_IMAGE_DIR="${FIRMWARE_IMAGE_DIR}-${BOARD_REVISION}"
    fi
 
-   if ! command -v vl805 > /dev/null; then
```
lurch: Would it be appropriate to add this check to _just_ the places in the code where the `vl805` program actually gets executed?

PR #98, opened 2020-02-28
Add pciutils dependency

XECDesign:


andrum99:
The package should probably also depend on libraspberrypi-bin, since it provides vxgencmd.

XECDesign:
That's already there - https://github.com/raspberrypi/rpi-eeprom/blob/debian/buster/debian/control#L13

andrum99:
> That's already there - https://github.com/raspberrypi/rpi-eeprom/blob/debian/buster/debian/control#L13

Sorry - my mistake. As you were.

PR #99, opened 2020-03-06
Allow for run from local git checkout

rfinnie:
This patch allows `rpi-eeprom-update` to fall back to the firmware/ directory in the same directory as `rpi-eeprom-update`, if all else fails.  Good for running directly from the git checkout, as the repo doesn't have an install target by itself and mostly assumes being part of a .deb.

### DIFF
--- a/rpi-eeprom-update
+++ b/rpi-eeprom-update
@@ -10,7 +10,19 @@ if [ -f /etc/default/rpi-eeprom-update ]; then
    . /etc/default/rpi-eeprom-update
 fi
 
-FIRMWARE_ROOT=${FIRMWARE_ROOT:-/lib/firmware/raspberrypi/bootloader}
+LOCAL_MODE=0
+if [ -n "$FIRMWARE_ROOT" ]; then
+   # Provided by environment
+   true
+elif [ -d /lib/firmware/raspberrypi/bootloader ]; then
+   # Default firmware root exists
+   FIRMWARE_ROOT=/lib/firmware/raspberrypi/bootloader
+else
+   # Work from local git checkout
+   LOCAL_MODE=1
+   FIRMWARE_ROOT="${script_dir}/firmware"
+fi
+
 # May be used to select beta or stable releases instead of the default critical updates.
 FIRMWARE_RELEASE_STATUS=${FIRMWARE_RELEASE_STATUS:-critical}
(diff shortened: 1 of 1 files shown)


timg236:
Looks good. Do you find that you need to run with -i to ignore the checksums ?

rfinnie:
> Looks good. Do you find that you need to run with -i to ignore the checksums ?

Actually no, I'm running on an OS installation where rpi-eeprom-images isn't even installed, so $checksums_file doesn't exist and the check is bypassed.

Nonetheless, do you think maybe bypass IGNORE_DPKG_CHECKSUMS when in this "local mode"?  Something like:
```diff
diff --git a/rpi-eeprom-update b/rpi-eeprom-update
index 03293f5..6d78b00 100755
--- a/rpi-eeprom-update
+++ b/rpi-eeprom-update
@@ -10,6 +10,7 @@ if [ -f /etc/default/rpi-eeprom-update ]; then
    . /etc/default/rpi-eeprom-update
 fi
 
+LOCAL_MODE=0
 if [ -n "$FIRMWARE_ROOT" ]; then
    # Provided by environment
    true
@@ -18,6 +19,7 @@ elif [ -d /lib/firmware/raspberrypi/bootloader ]; then
    FIRMWARE_ROOT=/lib/firmware/raspberrypi/bootloader
 else
    # Work from local git checkout
+   LOCAL_MODE=1
    FIRMWARE_ROOT="${script_dir}/firmware"
 fi
 
@@ -608,7 +610,7 @@ AUTO_UPDATE_BOOTLOADER=0
 AUTO_UPDATE_VL805=0
 MACHINE_OUTPUT=""
 JSON_OUTPUT="no"
-IGNORE_DPKG_CHECKSUMS=0
+IGNORE_DPKG_CHECKSUMS=$LOCAL_MODE
 
 while getopts A:adhif:m:ju:r option; do
    case "${option}" in
```

timg236:
Yes that looks good. Thanks

timg236:
I'll merge this on Monday when I get chance to test it for real but from code inspection it all looks good. Thanks for adding this.

rfinnie:
Thanks.  I've rebased the IGNORE_DPKG_CHECKSUMS addition so you should be good to review/merge.

timg236:
@rfinnie This change should now be available via the APT package. Thanks!

PR #100, opened 2020-03-12
pieeprom-2020-03-04: Netboot enhancements, self update and HDMI diagnostics

timg236:


timg236:
@XECDesign - Please could you update the Debian package when you get chance ? It only changes the beta so there's no need to go via the internal repo.

XECDesign:
Should be public within an hour.

timg236:
Thanks!

PR #107, opened 2020-04-07
Promote pieeprom-2020-03-19 to a stable update

timg236:
@XECDesign Nobody has reported any regressions with this beta so I think it's safe to go to stable once the corresponding documentation PR is merged.
Hopefully, this can be promoted to critical to get us out of diagnosing flashing error patterns but it's probably safer to go via a Pi Imager release first. Maybe we can put the stable release as a menu option in Pi Imager ?


XECDesign:
Alright, will do.

When you do the release on github, I can add it to Imager if @ghollingworth is alright with there being two recovery entries.

ghollingworth:
I'd rather not have two separate entries, it's another option for people to choose over (and we have no additional information which will help them make that decision).  I would instead suggest just switching that one when you've got something.  Since almost everyone gets their updates through linux I doubt it'll get used much anyway!

PR #110, opened 2020-04-12
recovery.bin: Update beta/stable image recovery image

timg236:
* Fix uart_2ndstage in the recovery.bin
* Fix missing watchdog reset if only the VL805 image is updated.
* Fix detection of VL805 EEPROM in recovery mode.
* If no EEPROM images are present then recovery.bin now renames
  itself and reboots instead of treating this as an error.


timg236:
@XECDesign fixed some fallout related to the VL805 changes in the stable/beta recovery.bin. The production version and recovery.bin used for rescue mode or critical update does not have these bugs.

Issue #113, opened 2020-04-16
Last beta/pieeprom-2020-16-04 doesn't follow the date format YYYY-MM-DD


txenoo:
All pieprom files have been following the YYYY-MM-DD format but last one 'pieeprom-2020-16-04.bin'  is using  YYYY-DD-MM, when next pieeeprom file will be released it would be more difficult to detect the last one as pieeprom-2020-16-04 would be the last in order during 2020.

$dpkg -s rpi-eeprom-images 
Package: rpi-eeprom-images
Status: install ok installed
Priority: optional
Section: misc
Installed-Size: 10669
Maintainer: Serge Schneider <serge@raspberrypi.org>
Architecture: all
Source: rpi-eeprom
Version: 5.6-1
Description: Raspberry Pi 4 boot EEPROM images
 This package contains the bootloader EEPROM images for Raspberry Pi 4
Homepage: https://github.com/raspberrypi/rpi-eeprom/

$ ls /lib/firmware/raspberrypi/bootloader/beta/pieeprom-*|tail -n 4
/lib/firmware/raspberrypi/bootloader/beta/pieeprom-2020-03-16.bin
/lib/firmware/raspberrypi/bootloader/beta/pieeprom-2020-03-19.bin
/lib/firmware/raspberrypi/bootloader/beta/pieeprom-2020-04-09.bin
/lib/firmware/raspberrypi/bootloader/beta/pieeprom-2020-16-04.bin

timg236:
Weird,  I must have been working late to slip into a different timezone. Will fix.

timg236:
@XECDesign sorry - renamed it now

lurch:
Fixed by #114

timg236:
Thanks @lurch

PR #114, opened 2020-04-17
Debian/buster

XECDesign:


lurch:
Fixes #113

lurch:
@XECDesign Sorry for not thinking of this earlier, but should there be a post-install rule to delete the `pieeprom-2020-16-04.bin` file (if it exists) for the benefit of people who installed the previous version of the package? There's no "month 16" so there's no danger of that ever being a "valid" filename.

XECDesign:
Do you mean from /boot or /lib? If from /boot, why? If from /lib, then dpkg removes previous files before updating a package, so it wouldn't be left behind.

lurch:
> dpkg removes previous files before updating a package, so it wouldn't be left behind.

Ahh, didn't realise that! Everything's okay then :+1:

PR #117, opened 2020-04-28
Promote 2020-04-16 to be the new default release as a critical update

timg236:
@XECDesign


lurch:
@JamesH65 I guess this implies that the documentation needs to be updated to reflect the fact that network boot on the Pi4 is no longer a "beta feature" ?

JamesH65:
I guess so, bootlfow_2711.md, plus there is a broken link on that page I just found that i will fix up at the same time.

JamesH65:
OK done and merged to master.

timg236:
Thanks!

PR #146, opened 2020-06-04
rpi-eeprom-update: compare versions algebraically

mackyle:
Use `-lt` rather than `\<` so that, for example, this:

    [ 2 -lt 17 ]

is true (whereas with `\<` in place of `-lt` it's false).

Signed-off-by: Kyle J. McKay <mackyle@gmail.com>

### DIFF
--- a/rpi-eeprom-update
+++ b/rpi-eeprom-update
@@ -272,7 +272,7 @@ checkDependencies() {
    BOARD_TYPE=$(((0x$BOARD_INFO >> 4) & 0xff))
    BOARD_REVISION=$((0x$BOARD_INFO & 0xf))
 
-   if [ ${BOARD_TYPE} = 17 ] && [ ${BOARD_REVISION} \< 4 ]; then
+   if [ ${BOARD_TYPE} -eq 17 ] && [ ${BOARD_REVISION} -lt 4 ]; then
       echo "Dedicated VL805 EEPROM detected"
       HAVE_VL805_EEPROM=1
    else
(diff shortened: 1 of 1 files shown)


lurch:
Does `-lt` definitely work for comparing hex values and not just decimal values?

MilhouseVH:
No.

rpi-eeprom-update
```diff
@@ -539,7 +539,7 @@ lookupVersionInfo()
    if [ "${HAVE_VL805_EEPROM}" = 1 ]; then
       getVL805UpdateVersion
       if [ -n "${VL805_CURRENT_VERSION}" ] && [ -n "${VL805_UPDATE_VERSION}" ]; then
-         if [ "${VL805_CURRENT_VERSION}" \< "${VL805_UPDATE_VERSION}" ];  then
+         if [ "${VL805_CURRENT_VERSION}" -lt "${VL805_UPDATE_VERSION}" ];  then
```
MilhouseVH, line 542: `-lt` compares integers, and will fail for a non-numeric version such as that used by VL805 which is correctly using `\<`.

rpi-eeprom-update
```diff
@@ -272,7 +272,7 @@ checkDependencies() {
    BOARD_TYPE=$(((0x$BOARD_INFO >> 4) & 0xff))
    BOARD_REVISION=$((0x$BOARD_INFO & 0xf))
 
-   if [ ${BOARD_TYPE} = 17 ] && [ ${BOARD_REVISION} \< 4 ]; then
+   if [ ${BOARD_TYPE} = 17 ] && [ ${BOARD_REVISION} -lt 4 ]; then
```
MilhouseVH, line 275: Strictly speaking, this should be
```
if [ ${BOARD_TYPE} -eq 17 ] && [ ${BOARD_REVISION} -lt 4 ]; then
```

rpi-eeprom-update
timg236: Maybe the VL805 version string could be converted to an integer via $((0x.

mackyle:
Updated patch with feedback from comments:

  * use 0x trick to compare VL805 versions
  * change `=` to `-lt` to be technically correct

timg236:
Thanks, will try to give it a test today

timg236:
Tested on Pi 4 4GB and Pi 4 8GB. Thanks for the fix!

Issue #151, opened 2020-06-06
Bootloader can't boot via USB-HDD after system reboot 


wildrun0:
After rebooting (sudo reboot), im getting stuck in bootloader with this loop:
```
Failed to open device: 'sdcard' (cmd 371a0010 status 1fff0001)
Failed to open device: 'sdcard' (cmd 371a0010 status 1fff0001)
root HUB port 1 init
root HUB port 3 init
USB MSD timed out after 20 seconds
```
If at this moment i replug usb-hdd, it will boot correctly, And sometimes it says that: 'SD: card detected' but there isnt no sd card in rpi.

Bootloader version:
```
pi@raspberrypi:~ $ vcgencmd bootloader_version
Jun  3 2020 13:53:47
version b5de8c32f4f45a12a1fdfe107254df82965f9d56 (release)
timestamp 1591188827
```
Bootloader configuration:
```
pi@raspberrypi:~ $ vcgencmd bootloader_config
[all]
BOOT_UART=0
WAKE_ON_GPIO=1
POWER_OFF_ON_HALT=0
DHCP_TIMEOUT=45000
DHCP_REQ_TIMEOUT=4000
TFTP_FILE_TIMEOUT=30000
ENABLE_SELF_UPDATE=1
DISABLE_HDMI=0
SD_BOOT_MAX_RETRIES=1
USB_MSD_BOOT_MAX_RETRIES=1
BOOT_ORDER=0xf41
```

lsusb -vvv output:
```
pi@raspberrypi:~ $ sudo lsusb -vvv

Bus 002 Device 002: ID 152d:0578 JMicron Technology Corp. / JMicron USA Technology Corp. JMS567 SATA 6Gb/s bridge
Device Descriptor:
  bLength                18
  bDescriptorType         1
  bcdUSB               3.00
  bDeviceClass            0
  bDeviceSubClass         0
  bDeviceProtocol         0
  bMaxPacketSize0         9
  idVendor           0x152d JMicron Technology Corp. / JMicron USA Technology Corp.
  idProduct          0x0578 JMS567 SATA 6Gb/s bridge
  bcdDevice           32.02
  iManufacturer           1 JMicron
  iProduct                2 JMS579
  iSerial                 3 6640391D
  bNumConfigurations      1
  Configuration Descriptor:
    bLength                 9
    bDescriptorType         2
    wTotalLength       0x0079
    bNumInterfaces          1
    bConfigurationValue     1
    iConfiguration          0
[...]

Issue #144, opened 2020-06-03
Kernel warning and network failure when attempting to use the network after bootloader times out.


dickontoo:
If the Pi 4 is booted with a boot order which includes the network before the ultimately successful boot device, but no network cable is plugged in, the kernel will crash when attempting to use it, once the link is restored.  I am using NFS boot, but I don't believe that is related.  On a control-alt-delete reboot, the bootloader no longer displays the diagnostics screen.

Kernel version:
[    0.000000] Linux version 5.4.42-v7l+ (dom@buildbot) (gcc version 4.9.3 (crosstool-NG crosstool-ng-1.22.0-88-g8460611)) #1319 SMP Wed May 20 14:12:03 BST 2020

**To Reproduce:**

Set BOOT_ORDER to, say, 0xf124.  Remove all USB bootable devices.  Install a bootable uSD card.  Remove the network cable.  Boot the Pi.

The bootloader will time out waiting for the USB and network stages, then fall back to the uSD card.  At this point, the kernel should boot.  With an NFS root, it will wait undefinitely for a cable.  Plugging the cable in at any time after the network boot timeout, including during the rainbow screen, will cause the kernel to crash when it attempts to send a packet.

**Expected behaviour**

The kernel run as usual.

**Screenshots**

Backtrace:
```
[    0.000000] Kernel command line: coherent_pool=1M 8250.nr_uarts=1 snd_bcm2835.enable_compat_alsa=0 snd_bcm2835.enable_hdmi=1 snd_bcm2835.enable_headphones=1 
video=HDMI-A-1:3840x2160M@60,margin_left=48,margin_right=48,margin_top=48,margin
_bottom=48 smsc95xx.macaddr=DC:A6:32:03:0F:D3 vc_mem.mem_base=0x3ec00000 vc_mem.mem_size=0x40000000  dwc_otg.lpm_enable=0 console=tty1 console=ttyS0,115200 root=/dev/nfs elevator=deadline rootwait rw nfsroot=172.29.23.1:/var/local/nfsroot/telly-buster ip=dhcp consoleblank=0 net.ifnames=0 netbooted
[...]
[    2.413704] bcmgenet fd580000.ethernet: failed to get enet clock
[    2.413788] bcmgenet fd580000.ethernet: GENET 5.0 EPHY: 0x0000
[    2.413869] bcmgenet fd580000.ethernet: failed to get enet-wol clock
[...]

Issue #58, opened 2019-11-22
No DHCPACK with DHCP relay agent


marcolefo:
Hi

I am trying to netboot my pi4 with no success.

After plug the pi, the green LED is fix. I saw DHCP requests on my server.
The Pi keep requesting DHCP server and the server answer with the IP.
I only see DHCPDISCOVER and DHCPOFFER. No DHCPACK.

I have tried with pieeprom-2019-11-18.bin and pieeprom-2019-10-16.bin.

My dhcp server is a isc-dhcp on a Debian Jessie. Note that it works for a pi3.
The definition for the subnet is :
```
subnet 10.10.10.0 netmask 255.255.255.0 {
        interface eth0;
        allow unknown-clients;
        range 10.10.10.230 10.10.10.234;
        option subnet-mask 255.255.255.0;
        option routers 10.10.10.254;

        next-server 10.10.10.1;
        option tftp-server-name "10.10.10.1";
}
```

Muxxxi:
Hi

I don't know if it helps, but the following configuration works for me so far.
I'm running the Docker Image `networkboot/dhcpd`.

```
subnet 192.168.1.0 netmask 255.255.255.0 {
        range 192.168.1.20 192.168.1.100;
      	next-server 192.168.1.1;
        option tftp-server-name "192.168.1.1";
    	always-broadcast false;
    	option vendor-class-identifier "PXEClient";
        option vendor-encapsulated-options "Raspberry Pi Boot";
        option routers 192.168.1.1;
        option domain-name-servers 8.8.8.8;
}
```

timg236:
Please can you post wireshark capture. Might be expecting SI_ADDR but won’t be able to look at this for a few days.

marcolefo:
> 
> 
> Hi
> 
> I don't know if it helps, but the following configuration works for me so far.
> I'm running the Docker Image `networkboot/dhcpd`.
> 
> ```
> subnet 192.168.1.0 netmask 255.255.255.0 {
>         range 192.168.1.20 192.168.1.100;
>       	next-server 192.168.1.1;
>         option tftp-server-name "192.168.1.1";
>     	always-broadcast false;
>     	option vendor-class-identifier "PXEClient";
>         option vendor-encapsulated-options "Raspberry Pi Boot";
>         option routers 192.168.1.1;
>         option domain-name-servers 8.8.8.8;
> }
> ```

No change with adding these options

marcolefo:
> 
> 
[...]

PR #154, opened 2020-06-12
pieeprom-2020-06-12 Improve support for powered USB SATA devices - BETA

timg236:
* Reset Ethernet MAC + PHY if final boot mode is not network boot
  See: Kernel warning and network failure when attempting to use
       the network after bootloader times out. #144
* Improve handling of multiple bootable USB devices and remove USB_MSD_BOOT_MAX_RETRIES
* Resolve: No DHCPACK with DHCP relay agent #58
* Toggle USB root hub port power for 200ms on the first USB MSD boot attempt
  See: Bootloader can't boot via USB-HDD after system reboot #151
* Update bootloader handover to support uart_2ndstage - requires a newer
  start.elf firmware which will be via rpi-update.
* Assert PCIe fundamental reset if the final bootmode was not USB-MSD
  because the OS might not do this before starting XHCI.


timg236:
@XECDesign - please could you generate a new release?

lurch:
> remove USB_MSD_BOOT_MAX_RETRIES

Oh, I guess that ought to be removed from the [documentation](https://github.com/raspberrypi/documentation/blob/master/hardware/raspberrypi/bcm2711_bootloader_config.md) then? (just out of curiosity, why was it removed?)

timg236:
It's removed because unlike SD where there is just one device to try there is an unknown and possibly changing number of bootable devices so the bootloader needs to try them all until a timeout occurs

PR #159, opened 2020-06-17
Debian/buster

XECDesign:


timg236:
@spl237 @ghollingworth 

There is now a stable-ish release for USB MSD boot so it should be ok to enable the raspi-config menu options. Although, the boot-order changes should really be gated or at least documented to require the stable release.

PR #164, opened 2020-06-22
Support python3 in the debian package

RaspberryPiFan:
This allows installing and building this on Ubuntu 20.04


timg236:
@XECDesign seems ok to me since this is a packaging change rather than a script change

XECDesign:
This change can lead to a broken install. If you only have python 3 and not python 2,  the shebang will point to a non-existing file and this change allows for this situation.

PR #169, opened 2020-06-24
Query checksums from rpi-eeprom instead of rpi-eeprom-images

waveform80:
I've been putting together an Ubuntu package for rpi-eeprom and noticed in rpi-eeprom-update that it queries the now deprecated rpi-eeprom-images package, instead of rpi-eeprom. I'm assuming this needs correcting now that rpi-eeprom-images is effectively empty?


timg236:
LGTM @XECDesign

Issue #172, opened 2020-06-28
Does not boot 2020-05-27-raspios-buster-armhf.zip


ilg-ul:
**Describe the bug**

The June 15 bootloader does not boot the 2020-05-27 **32-bit** image (but the 64-bit image is ok).

**To Reproduce**

- copy the `2020-05-27-raspios-buster-armhf.zip` to any USB stick known to work (I used a SanDisk Ultra USB 3.0 64 GB with balenaEtcher)
- plug the USB stick into RPi 4 USB 3.0 lower port
- power on

**Expected behaviour**

- the system boots and starts install

**Screenshots**

```
bootloader: c302dea0 Jun 15 2020

Read config.txt bytes 1784 ...
recovery4.elf not found (6)
Read start4.elf bytes 2252416 ...
Read fixup4.dat bytes 5400 ...
0x00c03112 0x0 0x0f
start4.elf: is not compatible
USB-MSD boot requires newer software
Get the latest software from https://www.raspberrypi.org/gownloads
ERROR: 00000044
```

**Bootloader version and configuration**

```
$ sudo vcgencmd bootloader_version
Jun 15 2020 14:36:19
version c302dea096cc79f102cec12aeeb51abf392bd781 (release)
timestamp 1592228179
```

**Additional context**

Exactly the same system with the same SanDisk USB stick boots and installs properly the 64-bit image `2020-05-27-raspios-buster-arm64.zip`.

I saw several messages in the forum suggesting all kind of wizardry, like overwriting files, but I'm not sure this is the right way, in this case the boot image is the official Raspberry Pi OS .zip file, normally it should be consistent.

I friend of mine had the same experience, so I guess the issue can be reproduced easily.

lurch:
This is because to be able to USB-boot you need both a new-enough bootloader (in the EEPROM) **and** a new-enough `start4.elf` (on the USB stick).
The `start4.elf` in the (stable) 32-bit Raspberry Pi OS isn't new enough to USB-boot (which is why you get the "start4.elf: is not compatible" message), but the `start4.elf` in the (beta) 64-bit Raspberry Pi OS _is_ new enough to USB-boot.

The documentation at https://www.raspberrypi.org/documentation/hardware/raspberrypi/bcm2711_bootloader_config.md explains what you need to do. (basically you need to copy the *.elf and *.dat files from https://github.com/raspberrypi/firmware/tree/master/boot to your USB stick)

ilg-ul:
[...]

PR #173, opened 2020-06-29
firmware/release-notes.md: add notice for 32-bit images

ilg-ul:
As a conclusion of #172, I added a small notice to the release notes, to document the solution.

I hope that this will save some time for those who try to use the new USB bootloader.

### DIFF
--- a/firmware/release-notes.md
+++ b/firmware/release-notes.md
@@ -4,6 +4,12 @@
    * Promote the latest beta EEPROM and recovery.bin to stable and
      feature freeze USB MSD support until a production release is ready.
 
+Please note that the recent 64-bit beta image
+`2020-05-27-raspios-buster-arm64.zip` includes the updated firmware
+and boots just fine, but the 32-bit stable image uses the older firmware
+and requires manual updates for the `.elf` and `.dat` files, as documented
+in the [#172 issue](https://github.com/raspberrypi/rpi-eeprom/issues/172).
+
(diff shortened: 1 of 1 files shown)


lurch:
Feels to me like this is probably information that belongs at https://github.com/raspberrypi/documentation rather than at https://github.com/raspberrypi/rpi-eeprom ? :shrug:  But it's @timg236 's call :smiley:

timg236:
> Feels to me like this is probably information that belongs at https://github.com/raspberrypi/documentation rather than at https://github.com/raspberrypi/rpi-eeprom ? 🤷 But it's @timg236 's call 😃

releases.md is linked from the downloads page and needs to be updated when a release is made. Keeping it here allows this to be updated in the same commit or at least avoids going via copy edit every time a link is made. 

Localizing this here would make the problem worse, however, localized text in the docs repo linked to from this page might be ok if someone to create that.

timg236:
Remove comment for the wrong bug :)

I think it would be best to add the text here but with a link from the release notes
https://github.com/raspberrypi/documentation/blob/master/hardware/raspberrypi/bcm2711_bootloader_config.md

ilg-ul:
Please feel free to adjust the text to better suit your needs, or to add links wherever you need.

Of course a better place would be the documentation project, but there the information is spread across multiple pages and I do not have time to work on this now.

You can very well close this PR if not needed, I think you got the point.

timg236:
Closing, the OS's have moved on

PR #175, opened 2020-07-06
Use /usr/bin/env python to make rpi-eeprom-config more portable

golddranks:
This change makes it work (at least) with NixOS out-of-the box, given that Python is installed in the environment.

### DIFF
--- a/rpi-eeprom-config
+++ b/rpi-eeprom-config
@@ -1,4 +1,4 @@
-#!/usr/bin/python
+#!/usr/bin/env python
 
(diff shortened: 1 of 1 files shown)


timg236:
@XECDesign @MilhouseVH 

Is this ok?

XECDesign:
Yes, this would be okay for RaspiOS.

MilhouseVH:
This is also OK for LibreELEC.

timg236:
Thanks

PR #178, opened 2020-07-08
Debian/buster

XECDesign:


debian/control
```diff
@@ -2,7 +2,7 @@ Source: rpi-eeprom
 Section: misc
 Priority: optional
 Maintainer: Serge Schneider <serge@raspberrypi.org>
-Build-Depends: debhelper (>= 11), help2man, python-minimal | python3-minimal
+Build-Depends: debhelper (>= 11), help2man, python-minimal
```
AaronDewes, line 5: Why did you revert that?

andrum99: The optional dependency on python3 was added to support Ubuntu. There was also a change to use `/usr/bin/env python` to makes the package more portable, which hasn't been reverted 🤔 was there an issue with the optional dependency?

lurch: #164 says "This change can lead to a broken install."

The `/usr/bin/env` change was made separately in #175 :slightly_smiling_face:

andrum99: Ah I see. That makes sense. From my reading, `/usr/bin/env` simply scans the path and picks the first binary it comes to - it doesn't have any interaction with Debian's mechanism for setting defaults, so there's no way to tell it that `python3` should be used if the user asks for `python`.

Canonical can make their own package - they don't have anything for `rpi-eeprom` at the moment.

lurch: Yeah, there's no `#!/usr/bin/env python_or_python2_or_python3` nor `#!/usr/bin/env python3_or_python_or_python2` :wink:

lurch:
Just a heads-up that the change from #179 hasn't made it into _this_ PR (yet).

XECDesign:
The package based on this PR is already in the repo. I think #179 would need to be a part of the next release, which would be a separate PR.

andrum99:
> Just a heads-up that the change from #179 hasn't made it into _this_ PR (yet).

That doesn't really matter too much. #179 is purely to support third-party use of the package.

debian/control
AaronDewes: It is: #181 :wink:

PR #181, opened 2020-07-08
Hack: detect right python version

AaronDewes:
This is both valid shell and python, if you execute this script, it uses a shell to detect if `python` or `python3` is available, then executes itself with python.

### DIFF
--- a/rpi-eeprom-config
+++ b/rpi-eeprom-config
@@ -1,4 +1,7 @@
-#!/usr/bin/env python
+#!/usr/bin/sh
+
+'''which' python > /dev/null && exec python "$0" "$@" || exec python3 "$0" "$@"
+ '''
 
 # rpi-eeprom-config
 # Utility for reading and writing the configuration file in the


XECDesign:
That's interesting, but why? I'd just add a quilt patch to make it python3 and change the dependencies accordingly.

lurch:
> to detect if `python` or `python` is available

Typo? :laughing: 

This is a cute trick, but I agree with @XECDesign that for distros which (still) ship Python2, just use the Raspberry Pi OS package; for distros which are Python3-only, build your own explicitly python3-only package.

XECDesign:
I mean that we can modify out package to patch the file when it's getting built to just use python 3 on our distro as well. Then the 'upstream' file remains unchanged for libreelec and everybody is happy.

andrum99:
I must be missing something here - does `python` not run the system's default python version anyway?

XECDesign:
I don't know how it works everywhere else, but in the case of debian buster, `python` will always be python 2.x. If that was not the case, everything would be horribly broken, since you can't use python 2 and python 3 interchangeably.

andrum99:
> I don't know how it works everywhere else, but in the case of debian buster, `python` will always be python 2.x. If that was not the case, everything would be horribly broken, since you can't use python 2 and python 3 interchangeably.
> 
> Edit: I believe it's all based on this - https://www.python.org/dev/peps/pep-0394/
> 
> Edit2: some more history - https://lwn.net/Articles/780737/

Thanks - I did a quick check on 2 distros I have in VMs. Ubuntu 20.04 Desktop has no python installed by default, and offers two packages when I ask for `python`: `python3` and `python3-is-python`. Linux Mint 20 has python3 installed by default, but does not map it to `python`.

I will leave the question of whether to support 3rd-party distros in this repo up to those that make such decisions. 🙃

waveform80:
> That's interesting, but why? I'd just add a quilt patch to make it python3 and change the dependencies accordingly.

FYI, this is precisely what I'm doing with the Ubuntu version of the package which is currently undergoing review.

timg236:
Just a thought but maybe the code can be mostly module instantiated by a minimal python2 or Python3 wrappers and the rpi-eeprom-config becoming a shell script which picks either.  I Can’t really look at this right now though

timg236:
Just a thought but maybe the code can be mostly module instantiated by a minimal python2 or Python3 wrappers and the rpi-eeprom-config becoming a shell script which picks either.  I Can’t really look at this right now though

andrum99:
Libreelec should just symlink python to python3 - the code already works in both versions.

timg236:
Do you mean link Python3 to Python ? Or alternatively have a shell script which calls
python rpi-eeprom-config.py or similar

andrum99:
> Do you mean link Python3 to Python ? Or alternatively have a shell script which calls
> python rpi-eeprom-config.py or similar

I mean I don't thing Raspberry Pi should be using hacks to accomodate other distros when there are easy solutions available at their end. The use of quilt has already been suggested, and that's what Ubuntu are using. Symlinking python to python3 may or may not be possible, depending on how Kodi handles python compatibility. Ultimately I think it should be up to the package maintainers on each platform to select whether the script is run by python or python3.

andrum99:
Actually - can you not just change the script so it invokes python3 anyway? I think I'm right in saying that python3 is installed on Raspberry Pi OS by default.

timg236:
No, the point of this PR so that this tool can be available other OS's without forcing them to have Python3 present, all of this because of some rather broken logic by to break everything in the Python2 to Python3 upgrade.

lurch:
> Libreelec should just symlink python to python3 - the code already works in both versions.

That rather naively assumes that there aren't any _other_ python scripts on Libreelec which are Python2-only and assume that python == python2 ? :shrug: (I dunno, I've never used Libreelec or Kodi)

andrum99:
> > Libreelec should just symlink python to python3 - the code already works in both versions.
> 
> That rather naively assumes that there aren't any _other_ python scripts on Libreelec which are Python2-only and assume that python == python2 ? 🤷 (I dunno, I've never used Libreelec or Kodi)

It turns out LibreElec is still using Python 2 and doesn't have Python 3. They're targetting Python 3 for their Kodi 19-based release.

waveform80:
To my mind the upstream here has done precisely what they should, i.e. make the script itself compatible with py2 and py3, but leave it to any downstream packagers to decide how they wish to install the script (and its dependencies).

On Ubuntu (20.04 onwards specifically), as py3 is included by default but py2 isn't, and because we've decided to leave the meaning of /usr/bin/python to individual administrators (via installation of the python-is-python2 or python-is-python3 packages or indeed neither), it makes sense to package a tool like this (which is also intended for inclusion by default) to be reliant on py3. But that's just Ubuntu, and it's almost certainly not what's "right" for all distros.

Personally, I'd leave the shebang as it is and leave it up to downstream packagers to deal with as they see fit. The only thing I'd request is that the script is maintained in a fashion suitable for use with either py2 or py3 (as it currently is) to facilitate that.

timg236:
_Personally, I'd leave the shebang as it is and leave it up to downstream packagers to deal with as they see fit. The only thing I'd request is that the script is maintained in a fashion suitable for use with either py2 or py3 (as it currently is) to facilitate that._

The only change that I have planned is to extract the code into a module to make it easier to re-use but this file and the module will retain Py2 + Py3 compatibility and will not add dependencies on any non-core Python modules.

There's a small unit test in test/test-rpi-eeprom-config which can be extended and if I can figure out how enabled in Github CI

I'm not an expert (in-fact total novice) on creating distro packages but if there's common functionality that we can maintain here to make things easier for other distros then I'm happy to do that.

lurch:
Sounds like you might have done this "Python" thing before @waveform80 ? :laughing: :wave: 

> The only change that I have planned is to extract the code into a module to make it easier to re-use

You're already using the
```python
if __name__ == '__main__':
    main()
```
idiom, so I don't think there are any changes needed to make the current code usable as a module already?

MilhouseVH:
> It turns out LibreElec is still using Python 2 and doesn't have Python 3. They're targetting Python 3 for their Kodi 19-based release.

LibreELEC also includes a `python` sym link to the respective `python2` (LE9)/`python3` (LE10) for "backward compatability".

---
To clear up any confusion, I think LibreELEC can safely be removed from this particular discussion.

Hard coding a `python3` shebang in this repo would not be an issue for LibreELEC - patching it back to `python` in the Python2-based branch of LibreELEC would be a trivial fix considering `rpi-eeprom-config.py` is Python2/Python3 compatible.

A `python3` shebang will work in the Python3-based branch of LibreELEC without any additional patching. `python` also works because of the aforementioned sym link (ie. `python3-is-python`).

XECDesign:
Although it really makes absolutely no different for Buster, the issue will keep getting raised because python 2's demise has been widely reported for a long time now.

In the long term, it should be python 3, so maybe if we change it to that now, it's one less thing to worry about in the future and will cut down on the noise.

timg236:
@XECDesign I'm happy for you to change the dependency to Python 3 since LibreELEC has the python3-is-python symlink

timg236:
Closing PR now that we have settled on /usr/bin/env python and python3-is-python for LibreElec and adding dependency on Python3 in Raspberry Pi OS

lpla:
I just updated the package on Raspberry Pi OS and package still depends on `python-minimal` (Python 2). Also, I guess that once this dependency changes to `python3-minimal` and Python 2 gets uninstalled in some setups like mine, there won't be any `/usr/bin/env python` in those (there is no `python-is-python3` package in buster).

PR #191, opened 2020-07-24
Fix VL805 version detection on Raspberry Pi 4B 8Gb v1.4

ww898:


rpi-eeprom-update
```diff
@@ -285,7 +285,7 @@ checkDependencies() {
       exit ${EXIT_SUCCESS}
    fi
 
-   if [ ${BOARD_TYPE} -eq 17 ] && [ ${BOARD_REVISION} -lt 4 ]; then
+   if [ ${BOARD_TYPE} -eq 17 ] && [ ${BOARD_REVISION} -le 4 ]; then
```
timg236, line 288: This doesn't seem right

Pi 4 8GB is revision 1.4 and does NOT have a dedicated VLI EEPROM.

ww898: Hi, I have some Raspberry Pi 4B 8G. One was updated after the fix:

timg236: Please can you explain what you think this commit does? The Pi4 8GB has a single EEPROM containing both the bootloader and the VLI firmware. It's not possible to upgrade the firmware independently on the 8GB, and 4GB may change to use the same scheme in future to simplify updates.

lurch: @ww898 The `vl805` command is deprecated - you can check whether your firmware is up to date by running `sudo rpi-eeprom-update`.

ww898: @timg236 Well, I understand. You can decline the pull request. 

@lurch I'm using Ubuntu 20.04 and so I have no `rpi-eeprom-update` command by default. So, I have to use this repository to update a firmware. Can you recommend me the official PPA for aarch64?

waveform80: > @timg236 Well, I understand. You can decline the pull request.
> 
> @lurch I'm using Ubuntu 20.04 and so I have no `rpi-eeprom-update` command by default. So, I have to use this repository to update a firmware. Can you recommend me the official PPA for aarch64?

I'm in the process of getting rpi-eeprom (and the userland dependency) into Ubuntu 20.04 (20.10 initially then SRU'd to 20.04). The rpi-eeprom package is now available in -proposed for both Focal and Groovy, but the userland dependency was still undergoing review. That's basically complete now and the package should enter -proposed this week.

A build of the userland package can be found in my [userland PPA](https://launchpad.net/~waveform/+archive/ubuntu/userland) but it's only for Groovy (20.10), though you could just manually install that in Focal. Honestly though I'd just wait a day for everything to hit -proposed; it should migrate pretty shortly after that.

timg236: I think you are missing the point of the original change. There is no VL805 EEPROM on 8GB so the there is no point in the script prompting the user about the presence of vl805 binaries.

The reason that the vl805 version changed is that the latest bootloader binary changed the embedded vl805 binary. 

On Pi4 8GB the vl805 is nothing more than a complicated way of running lspci

PR #199, opened 2020-08-12
Use python3 in debian package

lpla:
Keeping compatibility with Python 2 shebang in master for any other distribution system like LibreELEC but forcing python3 in debian branch makes some space in Raspbian/Raspberry Pi OS when installing it packaged by avoiding Python 2 dependency.

Replaces python-minimal and python dependencies with Python 3 versions (which are already installed in Raspbian/Raspberry Pi OS).

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -2,7 +2,7 @@ Source: rpi-eeprom
 Section: misc
 Priority: optional
 Maintainer: Serge Schneider <serge@raspberrypi.org>
-Build-Depends: debhelper (>= 11), help2man, python-minimal
+Build-Depends: debhelper (>= 11), help2man, python3-minimal
 Standards-Version: 4.1.3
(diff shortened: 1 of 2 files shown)


timg236:
I think we've already been around this loop @XECDesign ?

lurch:
More details (from when this was discussed before) in #181

XECDesign:
Quilt version I promised a long time ago - https://github.com/raspberrypi/rpi-eeprom/pull/203

lpla:
Closing pull request as looks like it is fixed by @XECDesign in #203 . Thank you! I will test it once it gets into Raspberry Pi OS.

lpla:
I confirm that latest update allows me to uninstall `python2*` packages without any issue or dependency. Thank you!

PR #208, opened 2020-08-20
rpi-eeprom-update: Upstream kernel fix

hvenev:
Upstream kernels do not list the board revision in /proc/cpuinfo. Get it
from the device tree instead.


lurch:
Rather than only using one method or the other, perhaps it makes sense to try one method, and then fall back to the other if it doesn't work? :man_shrugging:

timg236:
It should be ok. Will include in the September release

PR #212, opened 2020-09-01
Fix #211 - hexdump error on older Pi models

### LINKED ISSUE #211: Update script syntax errors on older Pi models
Running the latest `rpi-eeprom-update` on older Raspberry Pis gives the error:

    /usr/bin/rpi-eeprom-update: 273: /usr/bin/rpi-eeprom-update: arithmetic expression: expecting ')': "(0x00*
    0f >> 23) & 1"

This is due to `hexdump` eliding zeros in its output. I have boards with a BOARD_INFO of `0x00000002` and `0x0000000f`. These get an asterisk and newline inserted by `hexdump` as above. All that's needed is to pass the `-v` flag to emit all output. Line 271 of `rpi-eeprom-update` is where this error occurs but I haven't checked if there are any other uses like this.

brookst:
Bug fix for Raspberry Pis with low BOARD_INFO numbers. Tell hexdump to display all input data with `-v` so that zeros don't get mangled.

### DIFF
--- a/rpi-eeprom-update
+++ b/rpi-eeprom-update
@@ -268,7 +268,7 @@ getBootloaderUpdateVersion() {
 }
 
 checkDependencies() {
-   BOARD_INFO="$(hexdump -e '1/1 "%02x"' /sys/firmware/devicetree/base/system/linux,revision)"
+   BOARD_INFO="$(hexdump -ve '1/1 "%02x"' /sys/firmware/devicetree/base/system/linux,revision)"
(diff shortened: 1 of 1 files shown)


lurch:
I guess this is a follow-on from #208

brookst:
Yes, I didn't look into the git blame since this seems like an issue a lot of people may hit and there were no results just searching for the error. Hopefully a fix can be pushed out relatively quickly, I'm not sure what the hotfix/release process is.

lurch:
You managed to catch this error fairly quickly - the updated release only went out earlier today #210

timg236:
@hvenev

timg236:
I've merged this rather than reverting the original patch @XECDesign can you update the package ?

PR #224, opened 2020-09-11
pieeprom-2020-09-03: Promote latest stable to default/critical

timg236:
Promote the latest stable release of the bootloader, VLI EEPROMs and recovery.bin to the default (critical update) release.


timg236:
Creating a PR for the next default production Bootloader/VLI EEPROMs release. Planning to release this on Monday unless any showstopper regressions are found e.g. boards which boot ok on 2020-04-16 but now fail.

@XECDesign @MilhouseVH @vianpl @popcornmix @pelwell @waveform80

PR #226, opened 2020-09-15
Fix package architecture

andrum99:
Canonical have discovered an issue with the rpi-eeprom package: the package architecture is set to `all` - it should be `armhf arm64` - see https://bugs.launchpad.net/ubuntu/+bug/1884748. This PR corrects the issue here, in the upstream repo.

Rationale: the package will only work on a Raspberry Pi 4, so suggesting it works on architectures the Pi 4 doesn't use is misleading.

N.B. I have no association with Canonical - I just spotted it on their bug tracker.

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -9,7 +9,7 @@ Vcs-Browser: https://github.com/raspberrypi/rpi-eeprom/
 Vcs-Git: https://github.com/raspberrypi/rpi-eeprom.git
 
 Package: rpi-eeprom
-Architecture: all
+Architecture: armhf arm64
 Depends: ${shlibs:Depends}, ${misc:Depends}, libraspberrypi-bin, python3,
  binutils, raspberrypi-bootloader (>= 1.20190819), pciutils
 Breaks: rpi-eeprom-images (<<7.2)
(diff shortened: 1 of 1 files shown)


lurch:
ping @XECDesign (although it's not immediately obvious if this is an Ubuntu-only bug or not)

XECDesign:
I think the logic was that people may want to use the scripts from the package on their x86 desktop to prepare a card or patch the eeprom config.

timg236:
rpi-eeprom-config is sometimes used on X86 along with the unit test script. Although, it probably doesn't need to installed as a package on X86 i.e. just git clone

timg236:
This package depends on the raspberrypi-bootloader (GPU firmware) which is ARM only so it makes sense for this to not armhf/arm64

andrum99:
> I think the logic was that people may want to use the scripts from the package on their x86 desktop to prepare a card or patch the eeprom config.

The rpi-eeprom package is not installable on Raspberry Pi OS on x86, although the fact it gives an error message perhaps suggests it is meant to. I hadn't considered rpi-eeprom-config could be used standalone on x86, but since the package already depends on packages which are arm-only, I think this fix was appropriate. The rest of the package is also non-functional on x86. If there's a need to support rpi-eeprom-config on x86 in future, then I would imagine the correct way to do this would be to split it off into its own package.

PR #229, opened 2020-09-24
Update rpi-eeprom-config to provide wrap simple interactive and non-interactive update opertions

timg236:
* Add a simple 'git commit' inspired --edit option which retrieves the current config, spawns default editor then merges the new config into the latest release and schedules the update.

* Add  "--apply" flag which is a non-interactive version of the above where the new config file is specified on the command line instead of spawning and editor

@jdb @pelwell Here's the 'porcelain' additions which I was talking about earlier, no rush to merge but if you could update and add feedback when you get chance that would be great. Hopefully, the error handling is now improved.

git clone -b rpi_eeprom_config_apply https://github.com/timg236/rpi-eeprom.git
sudo ./rpi-eeprom/test/install

### DIFF
--- a/rpi-eeprom-config
+++ b/rpi-eeprom-config
@@ -37,7 +37,15 @@ def exit_handler():
         tmp_image = '%s/pieeprom.upd' % TEMP_DIR
         if os.path.exists(tmp_image):
             os.remove(tmp_image)
+        tmp_conf = '%s/boot.conf' % TEMP_DIR
+        if os.path.exists(tmp_conf):
+            os.remove(tmp_conf)
         os.rmdir(TEMP_DIR)
+
+def create_tempdir():
+    global TEMP_DIR
+    if TEMP_DIR is None:
+        TEMP_DIR = tempfile.mkdtemp()
 
 def exit_error(msg):
     """
@@ -79,17 +87,40 @@ def apply_update(config):
     Applies the config file to the latest available EEPROM image and spawns
     rpi-eeprom-update to schedule the update at the next reboot.
     """
-    global TEMP_DIR
     latest = get_latest_eeprom()
-    TEMP_DIR = tempfile.mkdtemp()
+    create_tempdir()
     tmp_update = "%s/%s" % (TEMP_DIR, 'pieeprom.upd')
     image = BootloaderImage(latest, tmp_update)
     image.write(config)
+    config_str = open(config).read()
 
-    sys.stdout.write("Updating bootloader EEPROM\n image: %s\nconfig: %s\n" % (latest, config))
+    sys.stdout.write("Updating bootloader EEPROM\n image: %s\nconfig: %s\n%s\n" %
+                     (latest, config, config_str))
     args = ['sudo', 'rpi-eeprom-update', '-d', '-f', tmp_update]
     resp = shell_cmd(args)
     sys.stdout.write(resp)
+
+def edit_config():
+    """
+    Implements something like visudo for editing EEPROM configs.
+    """
+    if 'EDITOR' not in os.environ:
+        exit_error('EDITOR environment variable not defined')
+    create_tempdir()
+    current_config = read_current_config()
+    tmp_conf = "%s/boot.conf" % TEMP_DIR
+    out = open(tmp_conf, 'w')
+    out.write(current_config)
+    out.close()
+    cmd = "\'%s\' \'%s\'" % (os.environ['EDITOR'], tmp_conf)
+    result = os.system(cmd)
+    if result != 0:
+        exit_error("Aborting update because \'%s\' exited with code %d." % (cmd, result))
+
+    new_config = open(tmp_conf, 'r').read()
+    if len(new_config.splitlines()) < 2:
+        exit_error("Aborting update because \'%s\' appears to be empty." % tmp_conf)
+    apply_update(tmp_conf)
 
 def read_current_config():
     """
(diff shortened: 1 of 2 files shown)


rpi-eeprom-config
```diff
@@ -97,34 +213,85 @@ class BootloaderImage(object):
...
-    parser.add_argument('eeprom', help='Name of EEPROM file to use as input')
+    """
+    Utility for reading and writing the configuration file in the
+    Raspberry Pi 4 bootloader EEPROM image.
+    """
+    description = """\
+Bootloader EEPROM configuration tool for the Raspberry Pi 4.
+Operating modes:
+
+1. Output the current bootloader to configuration STDOUT if no arguments are
```
lurch, line 224: "bootloader to configuration STDOUT" -> "bootloader configuration to STDOUT" ?

rpi-eeprom-config
```diff
@@ -97,34 +213,85 @@ class BootloaderImage(object):
...
+4. Applies a given config file to the latest available EEPROM image and
+   invokes rpi-eeprom-update to update the bootloader when the system is
+   rebooted.
+
+   rpi-eeprom-config --apply boot.conf
+
+   The latest available image is determined by querying \'rpi-eeprom-update -l\'
+   and depends on the rpi-eeprom-update configuration.
+
+5. Launches the default text editor ($EDITOR) to edit the current EEPROM
```
lurch, line 251: Numbers 1 to 4 above include a usage-example, but number 5 doesn't?

rpi-eeprom-config
```diff
@@ -97,34 +213,85 @@ class BootloaderImage(object):
...
+   configuration and then invokes rpi-eeprom-update to apply the updated
+   configuration to the latest available EEPROM image and schedule an update
```
lurch, line 253: The wording here seems to imply that rpi-eeprom-update **both** "applies the configuration" and "schedules the update", but AFACT it's rpi-eeprom-config itself that "applies the configuration", and rpi-eeprom-update only "schedules the update" ?

Combining it with the wording of Number 4 above, perhaps:
5. Launches the default text editor ($EDITOR) to edit the current EEPROM
   configuration, applies the updated configuration to the latest available
   EEPROM image and then invokes rpi-eeprom-update to update the
   bootloader when the system is rebooted.

lurch:
> Add a simple 'git commit' inspired --edit option which retrieves the current config, spawns default editor then merges the new config into the latest release and schedules the update.

I wonder if anybody ever might want to use the `--edit` option and **not** automatically bump to the latest eeprom version at the same time too? i.e. edit the config but keep the eeprom at the same version they're currently running. (IIRC `vcgencmd` can tell you which bootloader version you're currently running?)
Although off the top of my head I can't think of any compelling reasons why you might want to do that :wink:

rpi-eeprom-config
```diff
@@ -22,18 +27,128 @@ MAGIC_MASK = 0xfffff00f
...
 FILE_HDR_LEN = 20
 FILENAME_LEN = 12
+TEMP_DIR = None
+
+def exit_handler():
+    """
+    Delete any temporary files.
+    """
+    if TEMP_DIR is not None and os.path.exists(TEMP_DIR):
+        tmp_image = '%s/pieeprom.upd' % TEMP_DIR
```
lurch, line 37: Doesn't really make any difference, but personally I prefer doing things like this with `os.path.join(TEMP_DIR, 'pieeprom.upd')` rather than using string-formatting.

rpi-eeprom-config
```diff
@@ -22,18 +27,128 @@ MAGIC_MASK = 0xfffff00f
...
+        if os.path.exists(tmp_image):
+            os.remove(tmp_image)
+        tmp_conf = '%s/boot.conf' % TEMP_DIR
+        if os.path.exists(tmp_conf):
+            os.remove(tmp_conf)
+        os.rmdir(TEMP_DIR)
```
lurch, line 43: Alternatively, you could just use [shutil.rmtree](https://docs.python.org/3/library/shutil.html#shutil.rmtree) and skip the file-exists tests :wink:

rpi-eeprom-config
```diff
@@ -22,18 +27,128 @@ MAGIC_MASK = 0xfffff00f
...
+def shell_cmd(args):
+    start = time.time()
+    arg_str = ' '.join(args)
+    result = subprocess.Popen(args, stdout=subprocess.PIPE, stdin=subprocess.PIPE)
+
+    while time.time() - start < 5:
+        if result.poll() is not None:
+            break
+
+    if result.poll is None:
```
lurch, line 67: Should this be `result.poll()` ?

rpi-eeprom-config
timg236: That's deliberate. If the exists tests fails then it indicates a bug in a safer way than a random recursive directorty retre removal

rpi-eeprom-config
timg236: I'll think about it, os.path.join seems rather wordy

rpi-eeprom-config
timg236: good point will fix

rpi-eeprom-config
timg236: I decided to make the change in-case someone manages to sneak in a wierd path / extra separators

pelwell:
I can confirm that the latest update fixes the Python exception, but as discussed the error message from/triggered by rpi-eeprom-update is not getting through.

rpi-eeprom-config
```diff
@@ -239,23 +247,24 @@ Operating modes:
    The new image file can be installed via rpi-eeprom-update
    rpi-eeprom-update -d -f newimage.bin
 
-4. Applies a given config file to the latest available EEPROM image and
-   invokes rpi-eeprom-update to update the bootloader when the system is
-   rebooted.
+4. Applies a given config file an EEPROM image and invokes rpi-eeprom-update
```
lurch, line 250: "config file **to** an EEPROM image" ?

rpi-eeprom-config
```diff
@@ -97,34 +221,86 @@ class BootloaderImage(object):
...
+   specified OR the given output file if --out is specified.
+
+   rpi-eeprom-config [--out boot.conf]
+
+2. Extracts the configuration file from the given \'eeprom\' file and outputs
+   the result to STDOUT or the output file if --output is specified.
+
+   rpi-eeprom-config pieeprom.bin [--out boot.conf]
+
+3. Writes a new EEPROM image replacing the configuration file with the contents
```
andrum99, line 242: I would remvoe the word file to avoid confusion.

rpi-eeprom-config
timg236: done

jdb:
Hi Tim, i think you pinged the wrong jdb. Regards

timg236:
@pelwell Thanks for testing this. I'll merge to master and give it a couple of days before pushing the APT update to give people to shout if anything needs to be fixed for other distros

Issue #230, opened 2020-09-26
rpi4 AArch64 Arch Linux rpi-eeprom-update failure


FancyFane:
**Describe the bug**
When running the rpi-eeprom-update command it results in a failure on a fresh install of Arch Linux ARM 64 bit running on Raspberry Pi 4. 

**To Reproduce**
Install Arch Linux AArch64 version, selected support/troubleshooting packages

`# pacman -Syu --needed nfs-utils mkinitcpio-nfs-utils vim lsof tcpdump gnu-netcat sysstat bash-completion ethtool iputils man-pages man-db strace sudo rsync inetutils docker git mlocate ebtables socat cni-plugins python3 binutils fakeroot gcc pkgconf make autoconf automake lzo go go-tools python-docutils patch bison bc`

Install the yay utility here; then install the two packages needed for rpi-eeprom-update from the AUR

`$ yay -S rpi-eeprom raspberrypi-userland-aarch64-git`

Once this completed run the rpi-eeprom-update command to get the error below:

```
[root@alarm ~]# rpi-eeprom-update 
od: /sys/firmware/devicetree/base/system/linux,revision: No such file or directory
```

**Expected behaviour**
When working this should provide output to show if bootloader is up to date:

```
[root@alarm ~]# rpi-eeprom-update 
BCM2711 detected
VL805 firmware in bootloader EEPROM
BOOTLOADER: up-to-date
CURRENT: Thu Sep  3 12:11:43 UTC 2020 (1599135103)
 LATEST: Thu Sep  3 12:11:43 UTC 2020 (1599135103)
 FW DIR: /lib/firmware/raspberrypi/bootloader/stable
VL805: up-to-date
CURRENT: 00000000
 LATEST: 00000000
```

**Bootloader version and configuration**
```
[root@alarm ~]# vcgencmd bootloader_version
Sep  3 2020 13:11:43
version c305221a6d7e532693cc7ff57fddfc8649def167 (release)
timestamp 1599135103
```

**SD card boot (please complete the following information):**
```
[root@alarm ~]# uname -r
5.8.9-2-ARCH

[root@alarm ~]# cat /etc/os-release 
NAME="Arch Linux ARM"
PRETTY_NAME="Arch Linux ARM"
ID=archarm
ID_LIKE=arch
BUILD_ID=rolling
ANSI_COLOR="38;2;23;147;209"
HOME_URL="https://archlinuxarm.org/"
[...]

PR #231, opened 2020-09-27
Modifying BOARD_INFO var for Aarch64

FancyFane:
Fixes Issue #230

### DIFF
--- a/rpi-eeprom-update
+++ b/rpi-eeprom-update
@@ -268,7 +268,7 @@ getBootloaderUpdateVersion() {
 }
 
 checkDependencies() {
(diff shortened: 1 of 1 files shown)


lurch:
Somewhat ironically, this seems to be the inverse of #61 ?
See also #208 and #212 and #217 and #218 !!

This seems to be one of those "you can't please all of the people, all of the time" problems? :confused:

timg236:
If you want me to look at a commit then please write a proper commit message explaining what this change actually does.

As @luch says getting the board_info from OTP is the wrong approach see previous commits

PR #245, opened 2020-10-28
typos

lurch:
It's not clear what "does not have to be timestamp." is supposed to mean? :confused:


timg236:
You *could* invent your own monotonically incrementing number instead of UTC, but it's probably a bad idea :)

PR #249, opened 2020-11-02
WIP: Fully switch over to storing firmware under /lib instead of /var/lib

andrum99:
For some reason, the switch to /lib was only partially completed: this PR corrects that.


XECDesign:
I'm guessing the backups were kept in /var because they don't match the FHS description for /lib
```
The /lib directory contains those shared library images needed to boot the system and run the commands in the root filesystem, ie. by binaries in /bin and /sbin.
```
I'm not seeing a place backups fit neatly, but `/var/lib` seems to make the most sense. Of course Tim will remember why everything is the way it is and whether it's intentional much better than I would.

andrum99:
I assumed the ongoing usage of /var for backups was accidental, since there's also a backup directory under /lib, but maybe that's the mistake.

andrum99:
Per the FHS standard, rpi-eeprom should put the firmware images in /usr/lib. It's anyone's guess where the backup should go, but since it started out as a subdirectory of the same directory, that seems as good a place as any.

andrum99:
The FHS says /lib is only for stuff that the binaries in /bin and /sbin need to run, and neither script in this package gets installed into that directory - they both go into /usr/bin. I will leave it up to @timg236 to decide if that warrants action.

I'm happy that /var/lib is the probably the right place for the backup directory for the eeprom images, and withdraw this and #250.

lurch:
Also, looking at the script it seems it's only the firmware-configs that get backed up, not the firmware images themselves.
```bash
   # Backup the configuration of the currently loaded bootloader
   vcgencmd bootloader_config > "${TMP_EEPROM_CONFIG}"
   backup="${FIRMWARE_BACKUP_DIR}/pieeprom-backup-$(date +%Y%m%d-%H%M%S).conf"
   cp -f "${TMP_EEPROM_CONFIG}" "${backup}"
```

timg236:
Correct. There is no practical method for backing up the firmware. 

I think the config backup might get deleted at some point, it's never used and people can roll their own scripts to do that

lurch:
> I think the config backup might get deleted at some point, it's never used and people can roll their own scripts to do that

Maybe leave it in, but make it an off-by-default option? (i.e. something the user can enable with an extra command-line flag, if they wish)

PR #250, opened 2020-11-02
Packaging part of #249

andrum99:


andrum99:
Not sure if the entry in `rules` is actually needed.

(The bit in `rpi-eeprom.postrm`, which I have removed, is not - it cleans up `/lib/firmware/raspberrypi/bootloader` without needing a corresponding entry in this file).

debian/rpi-eeprom.postrm
```diff
@@ -20,9 +20,6 @@ set -e
 
 
 case "$1" in
-    purge)
-        rm -rf /var/lib/raspberrypi/bootloader/backup/
```
XECDesign, line 24: Could you elaborate on this change a bit? It still seems to create that directory here https://github.com/raspberrypi/rpi-eeprom/blob/0125155f9da31596252bbff4431d1bb7d8259933/rpi-eeprom-update#L112

andrum99: > Could you elaborate on this change a bit? It still seems to create that directory here
> 
> https://github.com/raspberrypi/rpi-eeprom/blob/0125155f9da31596252bbff4431d1bb7d8259933/rpi-eeprom-update#L112

I've probably done this the wrong way - the packaging change depends on #249, which corrects ${FIRMWARE_BACKUP_DIR} to be ${FIRMWARE_DIR}/backup, which ends up as `/lib/firmware/raspberrypi/firmware/backup`.

andrum99: Also see my comment above.

XECDesign: I meant why was the removed from purge, but you've answered that in the edit.

I think `purge` implies that you want it all gone. When `piuparts` runs, it checks that there is no cruft left over at all after a purge and this would fail that check. If the user wants to keep backups, they shouldn't purge them, right?

andrum99:
Withdrawn - see https://github.com/raspberrypi/rpi-eeprom/pull/249#issuecomment-720776740

PR #273, opened 2021-01-09
rpi-eeprom-update: Get bootloader build-timestamp from device-tree

timg236:


timg236:
/proc/device-tree/chosen/bootloader/version is populated in the latest firmware and should be used in preference to vcgencmd bootloader_versions because it does not depend on VCHI and is non-blocking.
There will be a separate PR to use a similar approach for the bootloader_config but that depends on the firmware  + kernel updates.

PR #283, opened 2021-01-21
rpi-eeprom-update: Remove deprecated flashrom script options.

timg236:
The muxing of the SPI EEPROM and analog audio pins makes flashrom
usage unreliable and having two flashing mechanisms adds unnecessary
complexity to this script.

The bootloader supports EEPROM updates from USB-MSD, Network
and rpiboot there is no real need to support flashrom within this
utility. Therefore, the option to use flashrom can be removed from this
script.


rpi-eeprom-update
```diff
@@ -230,39 +229,7 @@ applyUpdate() {
...
-   if [ -f "${VL805_UPDATE_IMAGE}" ]; then
-      echo "Applying VL805 image ${VL805_UPDATE_IMAGE}"
-      vl805 -w "${VL805_UPDATE_IMAGE}"
-   fi
-
-   echo "Applying bootloader update ${BOOTLOADER_UPDATE_IMAGE}"
+   applyRecoveryUpdate
 }
 
 # Use the version reported by the loaded EEPROM instead of attempting to retrieve
```
andrum99, line 235: The comment below this line refers to audio glitches when using the flashrom method - probably should be removed as well.

timg236:
Merging now, there will likely be a new APT update of a beta release on Friday.

N.B. Flashrom still works, it's just not being maintained in this script. If anyone wants to use flashrom then the approach would be to call this script or rpi-eeprom-config to from init to generate the pending update and run flashrom when it's safe to do so on that system. Previous versions of the source for calling flashrom can be copied and pasted from here :)

PR #287, opened 2021-01-26
rpi-eeprom-config: Properly decode configuration when extracted from …

vianpl:
…sysfs

The previous implementation was reading the sysfs file as plain text and
encoding it as 'ascii' to remove all the trailing zeros. This is wrong
twofold. To start with, the sysfs file we're querying is a binary
file[1], and we're reading it as a string. On top of that we're
benefiting that *some* python implementations of string.encode() will
deal with trailing zeros.

Fix this by treating the data as binary and decoding it as a string. On
top of that use 'utf-8.'

[1] sysfs files are generally text based, but there is also the option
to output binary data. Our configuration file does the later.


vianpl:
I need to reword the commit message

PR #288, opened 2021-01-26
rpi-eeprom-config: Properly decode sysfs binary files

vianpl:
The previous implementation was reading the sysfs files as plain text
and encoding them as 'ascii' to remove all the trailing zeros. This is
wrong twofold. To start with, the sysfs file we're querying are binary
files[1], and we're reading it as a string. On top of that we're
benefiting that *some* python implementations of string.encode() will
deal with trailing zeros.

Fix this by marking the files as binary and decoding them as strings
before consuming them.

[1] sysfs files are generally text based, but there is also the option
to output binary data. Our configuration file does the later.


vianpl:
Thanks!

PR #290, opened 2021-01-29
rpi-eeprom-update: Reference raspi-config from version message

timg236:
Update the human readable version of the version information to indicate
that raspi-config may be used to change the release type.
raspi-config can be replaced with other program names if necessary.


pelwell:
Looks good to me. I was wondering about whether there is any need to display `FIRMWARE_IMAGE_DIR` when `FIRMWARE_RELEASE_STATUS` is more user-friendly, but with a variety of options for `FIRMWARE_ROOT` it's probably no bad thing.

timg236:
Thanks @pelwell I'll give it another quick check on Pi400 before merging. The reason  for including the full path is that it's really easy to copy and paste the strings and run ls -l on it

PR #291, opened 2021-02-04
rpi-eeprom-config/update: Exit cleanly when not run on rpi4

vianpl:
Some users might forcibly install this on a board that isn't an rpi4.
Exit early, and explain why the program can't be run.

### DIFF
--- a/rpi-eeprom-config
+++ b/rpi-eeprom-config
@@ -32,6 +32,15 @@ FILE_HDR_LEN = 20
 FILENAME_LEN = 12
 TEMP_DIR = None
 
+def rpi4():
+    compatible_path = "/sys/firmware/devicetree/base/compatible"
+    if os.path.exists(compatible_path):
+        with open(compatible_path, "rb") as f:
+            compatible = f.read().decode('utf-8')
+            if "bcm2711" in compatible:
+                return True
+    return False
+
 def exit_handler():
     """
     Delete any temporary files.
@@ -327,9 +336,11 @@ images.
     parser.add_argument('eeprom', nargs='?', help='Name of EEPROM file to use as input')
     args = parser.parse_args()
 
-
     if (args.edit or args.apply is not None) and os.getuid() != 0:
         exit_error("--edit/--apply must be run as root")
+
+    if (args.edit or args.apply is not None) and not rpi4():
+        exit_error("--edit/--apply must run on a Raspberry Pi 4")
 
     if args.edit:
         edit_config(args.eeprom)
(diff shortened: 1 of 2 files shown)


rpi-eeprom-config
```diff
@@ -327,6 +336,8 @@ images.
     parser.add_argument('eeprom', nargs='?', help='Name of EEPROM file to use as input')
     args = parser.parse_args()
 
+    if not rpi4():
+        exit_error("This tool only works on a Raspberry Pi 4")
```
vianpl, line 340: To be more consistent I should maybe exit with no error? or the other way around, and have rpi-eeprom-update exit with an error if !rpi4.

timg236: I frequently use rpi-eeprom-config under Cygwin to modify the EEPROM config when using rpiboot/usbbot for CM4. A warning in the --edit/--apply modes which wrap rpi-eeprom-update might be useful though

vianpl: Oh! good point. I'll prepare a v2.

timg236: Thanks, there's a very minimal unit test file to verify the script level commands for patching pieeprom.bin
./test/test-rpi-eeprom-config

rpi-eeprom-update -h needs to work on any host so that help2man can be run on other platforms.

vianpl:
@timg236 I addressed your comments. Can be re-reviewed.

andrum99:
I'm wondering if the error message should mention Pi 400 as well. (Since the EEPROM tools are intended for use on Pi 4, Pi 400 and Compute Module 4. Compute Module 4 is for industrial users who should know about these things anyway).

vianpl:
> I'm wondering if the error message should mention Pi 400 as well. (Since the EEPROM tools are intended for use on Pi 4, Pi 400 and Compute Module 4. Compute Module 4 is for industrial users who should know about these things anyway).

Maybe it's too much of an assumption on my part, but I associate Raspberry Pi 4 without any suffixes with the extended family of boards (that is rpi4b, cm4, and pi400).

andrum99:
> > I'm wondering if the error message should mention Pi 400 as well. (Since the EEPROM tools are intended for use on Pi 4, Pi 400 and Compute Module 4. Compute Module 4 is for industrial users who should know about these things anyway).
> 
> Maybe it's too much of an assumption on my part, but I associate Raspberry Pi 4 without any suffixes with the extended family of boards (that is rpi4b, cm4, and pi400).

My question was aimed more at @timg236 😉

timg236:
I think until management decides upon a better term Pi4 is equivalent to Raspberry Pi BCM2711. If you want to be specific then use Pi4B, CM4, Pi400.
A BCM2711 without an EEPROM (e.g. if I boot from VC JTAG) would still be compatible with this software. You'd just get an empty string for the eeprom config and version of unknown/zero.

PR #298, opened 2021-02-18
rpi-eeprom-update: Use 'tr' instead of 'strings'

vianpl:
The 'strings' utility is installed by binutils whereas 'tr' belongs with
coreutils. Minimal systems will only contain the latter, due to binutils'
size (20-50 MB). So, convert all uses of 'strings' to 'tr' so as to
avoid the package dependency.

The second 'tr' usage converts non-ASCII characters into newlines so as
to isolate the "BUILD_TIMESTAMP=1234567890" statement (similar to what
'strings' does). There is no need for this in the first one: non-ASCII
characters are simply discarded as DT aliases have a fixed one line
format.

### DIFF
--- a/rpi-eeprom-update
+++ b/rpi-eeprom-update
@@ -113,7 +113,7 @@ getBootloaderConfig() {
    local blconfig_nvmem_path=""
 
    if [ -f "${blconfig_alias}" ]; then
-      local blconfig_ofnode_path="/sys/firmware/devicetree/base"$(strings "${blconfig_alias}")""
+      local blconfig_ofnode_path="/sys/firmware/devicetree/base"$(tr -cd [:print:] < "${blconfig_alias}")""
       local blconfig_ofnode_link=$(find -L /sys/bus/nvmem -samefile "${blconfig_ofnode_path}" 2>/dev/null)
 
       if [ -e "${blconfig_ofnode_link}" ]; then
@@ -288,7 +288,7 @@ getBootloaderUpdateVersion() {
    match=".*/pieeprom-[0-9][0-9][0-9][0-9]-[0-9][0-9]-[0-9][0-9].bin"
    latest="$(find "${FIRMWARE_IMAGE_DIR}/" -maxdepth 1 -type f -size "${EEPROM_SIZE}c" -regex "${match}" | sort -r | head -n1)"
    if [ -f "${latest}" ]; then
-      BOOTLOADER_UPDATE_VERSION=$(strings "${latest}" | grep BUILD_TIMESTAMP | sed 's/.*=//g')
+      BOOTLOADER_UPDATE_VERSION=$(tr -c [:print:] "\n" < "${latest}" | sed '/^BUILD_TIMESTAMP=/s/.*=//p;d')
       BOOTLOADER_UPDATE_IMAGE="${latest}"
    fi
 }


vianpl:
Note that with this the Debian control should be updated to lose the dependency.

rpi-eeprom-update
```diff
@@ -288,7 +288,7 @@ getBootloaderUpdateVersion() {
    match=".*/pieeprom-[0-9][0-9][0-9][0-9]-[0-9][0-9]-[0-9][0-9].bin"
    latest="$(find "${FIRMWARE_IMAGE_DIR}/" -maxdepth 1 -type f -size "${EEPROM_SIZE}c" -regex "${match}" | sort -r | head -n1)"
    if [ -f "${latest}" ]; then
-      BOOTLOADER_UPDATE_VERSION=$(strings "${latest}" | grep BUILD_TIMESTAMP | sed 's/.*=//g')
+      BOOTLOADER_UPDATE_VERSION=$(cat "${latest}" | tr -c [:print:] "\n" | grep BUILD_TIMESTAMP | sed 's/.*=//g')
```
ptesarik, line 291: What about saving some forks and execs?
* This `cat` is unnecessary. `tr` gets its input from standard input, so simply use shell redirection.
* `sed` can match by pattern, so you need no `grep`.

You may also want to be slightly more careful with the regexp:

```
       BOOTLOADER_UPDATE_VERSION=$(tr -c [:print:] "\n" < "${latest}" | sed '/^BUILD_TIMESTAMP=/s/.*=//p;d')
```

vianpl: Thanks for the comments, I'll test everything and update the patch.

timg236:
The change looks good to me, are you happy for this to be merged now?

vianpl:
@timg236 Yes, thanks!

timg236:
Could this be a non-standard version of tr?

https://www.raspberrypi.org/forums/viewtopic.php?p=1825133#p1825133

vianpl:
@timg236 I've been looking into it for a while and have no clue of what might be happening. My local RPi OS 64bit build doesn't seem to suffer from this, during or after the upgrade.

timg236:
Thanks for checking! I think there must be some unknown local difference but I can't imagine what could cause that.

pelwell:
There's an update from the user: https://www.raspberrypi.org/forums/viewtopic.php?p=1825520&sid=c2d94a624f16e56b3e0dabb63d063492#p1825520

vianpl:
@ptesarik You have more bash experience than me, does the 'corruption' the user is seeing here https://github.com/raspberrypi/rpi-eeprom/pull/298#issuecomment-784432732 ring a bell?

lurch:
It's not 'corruption' but `tr -cd [:print:] ` is somehow getting transformed to `tr -cd i n p r t ` - note that the second version has all the letters of 'print' but in alphabetical order and with a space between them.
I see that `rpi-eeprom-update` has `#!/bin/sh` as its interpreter, so maybe in the "not-working case" `/bin/sh` has been symlinked to a different shell, and this other shell is interpreting `[:print:]` (and converting it to `i n p r t`) rather than passing it to `tr` unmodified? Perhaps putting single-quotes around `'[:print:]'` is all that's needed to fix this problem?

lurch:
This is _hopefully_ now fixed by #302 :slightly_smiling_face:

pelwell:
An interesting observation from the user: https://www.raspberrypi.org/forums/viewtopic.php?p=1826238#p1826238

lurch:
> An interesting observation from the user: https://www.raspberrypi.org/forums/viewtopic.php?p=1826238#p1826238

The user's output there includes `/usr/bin/rpi-eeprom-update: 596: [: Illegal number:`

Line 596 of rpi-eeprom-update says `if [ "${BOOTLOADER_UPDATE_VERSION}" -gt "${BOOTLOADER_CURRENT_VERSION}" ]; then` so I guess "Illegal number" means that either `$BOOTLOADER_UPDATE_VERSION` or `$BOOTLOADER_CURRENT_VERSION` aren't getting set to a valid number (and thus causing the `-gt` comparison to fail). Those variables are set by the functions `getBootloaderCurrentVersion` and `getBootloaderUpdateVersion` which get called just before line 596.

I've got absolutely no idea why creating a directory named `n` in their root directory would cause things to go wrong!! But at least we've now confirmed that it's "user behaviour" that was causing `rpi-eeprom-update` to fail.

> one of my "customizations" is to add /{a,b,c,d,...z} to the root directory to be used for temporary mount points.

Maybe the user ought to be introduced to the `/mnt` directory, which is a much better place for creating temporary mountpoints :wink:

vianpl:
> An interesting observation from the user: https://www.raspberrypi.org/forums/viewtopic.php?p=1826238#p1826238

twilight

Tried to reproduce it here, but no luck.

PR #306, opened 2021-03-04
make-release: Add a script for generating Raspberry Pi Imager releases

timg236:
Create an EEPROM update zip files for use with the Raspberry Pi Imager.
The Raspberry Pi Imager JSON will soon be updated to support the 3
different boot priority choices offered by raspi-config.


timg236:
@XECDesign @peterharperuk 

Added a script to automated the creation of the zip files for RPI imager, please shout if you aren't happy with the file-names etc. I've verified that the files look sensible but not had a chance to boot each version yet.

peterharperuk (approved):
Looks good to me. Do we always use the latest "critical" binaries then rather than the latest stable release?

timg236:
> Looks good to me. Do we always use the latest "critical" binaries then rather than the latest stable release?

Yes. The default imager is always the default manufacture releases (which is updated for critical updates). There's no UI option to select the latest stable release. It might be useful but the main advantage of the Raspberry Pi Imager is that it's very simple to use so we've avoided adding too many options for the moment.

XECDesign:
LGTM

lurch:
This PR adds a `imager/recovery.bin` file, but that doesn't appear to be used by the script?

PR #308, opened 2021-03-05
Refactor the imager script to reduce duplication and remove reliance on symlinks

lurch:
Just a suggestion... feel free to chop'n'change as you see fit :slightly_smiling_face:

### DIFF
--- a/imager/make-imager-release
+++ b/imager/make-imager-release
@@ -1,0 +1,7 @@
+#!/bin/sh
+
+set -e
+
+script_dir=$(cd "$(dirname "$0")" && pwd)
+
+${script_dir}/make-release critical 2020-09-03 000138a1 "${script_dir}" release rpi-boot-eeprom-recovery
--- a/imager/make-release
+++ b/imager/make-release
@@ -1,11 +1,16 @@
 #!/bin/sh
 
 # Generates three variants of the rpi-eeprom-recovery.zip file for
-# SD, USB and NETWORK priority matching the raspi-config options.
+# SD, USB and NETWORK priority matching the raspi-config options,
+# plus a default (same as SD)
 
 set -e
 
+FIRMWARE_RELEASE="critical"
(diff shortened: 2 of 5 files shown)


timg236:
It's an interesting idea but the reason I didn't parameterize it is that the imager only supports the production release.

timg236:
Although thinking about this over some coffee, perhaps you could move this under a tools subdirectory and have a default configuration for 'build the default production release' and another option for build custom release.

I can this script being useful for someone who wants to take the latest stable release + chosen config to create a Pi updater SD-CARD e.g. network boot and latest EEPROM updater card for a classroom of Pis

lurch:
> the imager only supports the production release.

Yup, and that's why the `FIRMWARE_RELEASE="critical"` is hardcoded as a constant (rather than being a command-line parameter).
In effect, all this PR does is change the "external interface" from:
```
./make-release 2020-09-03-vl805-000138a1
```
to:
```
./make-release 2020-09-03 000138a1
```
and get rid of the symlinks in the `imager` directory in the process. The output files it creates are the same.

> I can this script being useful for someone who wants to take the latest stable release + chosen config to create a Pi updater SD-CARD e.g. network boot and latest EEPROM updater card for a classroom of Pis

I hadn't even considered that, but I guess that's a possibility :slightly_smiling_face:

timg236:
I was thinking that you could parameterize everything but have a trivial wrapper script to build the latest official release.

e.g. make-imager-release would just call make-release with the FW release dir, version, etc. We only change that when we want to bump the official release and anyone can re-run it without risking typing the wrong thing.

For custom releases, users can pass in whatever args they want or fork the repo with their alternate config files

lurch:
Done that now :slightly_smiling_face:

timg236:
Looks good, will give it a proper test on Monday and merge

timg236:
LGTM

PR #310, opened 2021-03-09
Add make-beta-release as a wrapper around make-release

lurch:
Builds `beta_release/*.zip` from the latest files found in `firmware/beta/`

### DIFF
--- a/imager/make-beta-release
+++ b/imager/make-beta-release
@@ -1,0 +1,13 @@
+#!/bin/sh
+
+set -e
+
+script_dir=$(cd "$(dirname "$0")" && pwd)
+
+firmware_status=beta
+firmware_dir=${script_dir}/../firmware/${firmware_status}
+pieeprom_version=$(basename $(ls ${firmware_dir}/pieeprom-*.bin | sort -V | tail -1) .bin | cut -d- -f2-5)
+vl805_version=$(basename $(ls ${firmware_dir}/vl805-*.bin | sort -V | tail -1) .bin | cut -d- -f2)
+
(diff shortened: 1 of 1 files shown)


timg236:
Perhaps the release name could be passed as a command-line argument so that this could be used with 'stable' or custom release names?

lurch:
It _could_, but I kinda assumed that people would only ever want to use the recovery-mode to flash eeprom images for either the critical firmware (which is what `make-imager-release` does) **or** the very latest beta firmware (which is what `make-beta-release` does) ? :man_shrugging: 

But I've just pushed a tweak which borrows the inherit-value-from-envvar approach used by `rpi-update` :slightly_smiling_face:

timg236:
> It _could_, but I kinda assumed that people would only ever want to use the recovery-mode to flash eeprom images for either the critical firmware (which is what `make-imager-release` does) **or** the very latest beta firmware (which is what `make-beta-release` does) ? 🤷‍♂️
> 
> But I've just pushed a tweak which borrows the inherit-value-from-envvar approach used by `rpi-update` 🙂

I think 'critical' release tends to be used by people who don't know about the bootloader and 'stable' is used by people who want new features which is quite common.

Now that the excitement over USB boot has faded away 'beta' is back to the original purpose i.e. support OS distro maintainers or to verify bug fixes

timg236:
Merging looks good and thanks for the additions. I think it will end up being folded into make-release with proper command line args. I only added the script in the first release because I knew that if I had to generate 3 different bootloader variants I'd get at least one of them wrong at some point.

However, reading back through forum posts I can see this being useful for resellers / education etc so we might want to direct people towards these scripts via docs/forums as a handy way of creating an SD card to customize 10s of Pi4s / Pi400s

PR #319, opened 2021-03-23
2021-03-18 - Promote 2021-03-18 to default/critical release

timg236:
Interesting changes since the last default release:-

   * Display VC_BUILD_ID strings instead of the SHA256 hash
   * Add support for [cm4] and [pi400] config conditionals filters.
   * Change network boot to use the same "RXID" Ethernet PHY configuration as the 5.10 kernel
   * TFTP - reply to duplicate ACKS
   * Skip rendering of HDMI diagnostics display for the first 8 seconds unless an error occurs.
   * UDP checksum fixes
   * Add support for the BCM2711 XHCI controller - BOOT_ORDER 0x5
   * XHCI protocol layer fixes for non-VLI controllers
   * Avoid USB MSD timeout of there is only one device
   * Implement tryboot for OS upgrade fallback
   * Check the update-timestamp before applying an update in SELF-UPDATE mode


timg236:
Preparing the default bootloader release update - needs final test and agreement before rolling this out.

timg236:
Pre-release for Raspberry Pi Imager zips
https://github.com/raspberrypi/rpi-eeprom/releases/tag/v2021.03.18-138a1

lurch:
"of there is" -> "if there is"

lurch:
To help people adapt from the old terminology to the new terminology, I wonder if it might be worth updating the header to `Promote 2021-03-18 from LATEST (STABLE) to DEFAULT (CRITICAL) - DEFAULT` ? :shrug:

timg236:
Merging - I'll do a few more smoke test to verify the MIN_VERSION functionality before requesting the APT package is updated

seamusdemora:
After a quick scan through this thread, I am under the impression that the critical/default eeprom firmware is now 2021-03-18. I've just done apt-get update && upgrade, but : 

```bash
$ sudo rpi-eeprom-update
BOOTLOADER: up-to-date
   CURRENT: Thu  3 Sep 12:11:43 UTC 2020 (1599135103)
    LATEST: Thu  3 Sep 12:11:43 UTC 2020 (1599135103)
   RELEASE: default (/lib/firmware/raspberrypi/bootloader/default)
            Use raspi-config to change the release.
```

Have I missed something? I thought eeprom updates were accomplished as part of the `upgrade` process...

lurch:
> Have I missed something?

The bit where Tim said "...before requesting the APT package is updated" ? :wink:

seamusdemora:
Well... I did say it was a "quick scan" - not an "accurate scan"    😎

lurch:
Oh, and note also that even when this _does_ get released as an APT package, a simple `apt upgrade` won't update you from 2020-09-03 because of #321  (a `sudo rpi-eeprom-update -a` would be needed instead).

seamusdemora:
@lurch : Thanks for the heads-up... isn't that in contravention to [the documentation](https://www.raspberrypi.org/documentation/hardware/raspberrypi/booteeprom.md)? I've read the following under `Automatic updates`

>The `rpi-eeprom-update` `systemd` service runs at startup and applies an update if a new image is available, automatically migrating the current bootloader configuration.  

I'm trying to resolve a troubling issue on my system: 

When `pieeprom-2020-04-16.bin` was in the `stable` folder, I set two 'configuration items': 
>WAKE_ON_GPIO=0
POWER_OFF_ON_HALT=1 

My purpose was to evaluate  a new **low power state** available & documented at the time. After making the changes, and testing, I found I could achieve approx **40mA current draw after `shutdown`**. This same RPi 4 now has this: 

```bash
$ vcgencmd bootloader_version
Sep  3 2020 13:11:43
version c305221a6d7e532693cc7ff57fddfc8649def167 (release)
timestamp 1599135103
update-time 0
capabilities 0x00000000
```

... and ...

```bash
$ vcgencmd bootloader_config
[all]
BOOT_UART=0
WAKE_ON_GPIO=0
POWER_OFF_ON_HALT=1
DHCP_TIMEOUT=45000
DHCP_REQ_TIMEOUT=4000
TFTP_FILE_TIMEOUT=30000
ENABLE_SELF_UPDATE=1
DISABLE_HDMI=0
BOOT_ORDER=0xf41
```

Here is the issue: 
So - `WAKE_ON_GPIO=0` and `POWER_OFF_ON_HALT=1` - just as it was when I set it in `pieeprom-2020-04-16.bin`. However, instead of the current draw decreasing from 500mA to 40mA, it now actually increases from 500mA to 620mA! 

After reading your post, I thought I would still be using `pieeprom-2020-04-16.bin`. I began checking & found that I (apparently) am now using the `Sep  3 2020` version; I also found that the low power state is not behaving as it did before. I didn't change my system bootloader config, nor did I manually install the `Sep  3 2020` bootloader. 

Can you explain - or offer any theories to explain what has happened?

lurch:
@seamusdemora Those are questions for @timg236 rather than me - I'm not directly involved with rpi-eeprom, I just try to help out with GH issues where I can. It sounds like you probably ought to open a new issue for your increased-power-in-halt problem?

seamusdemora:
OK - I'll do that. But to close out this issue, can you clarify the first question: 

Your comment:  
> simple `apt upgrade` won't update you from 2020-09-03 because of #321 (a `sudo rpi-eeprom-update -a` would be needed instead).  

My question: 
>isn't that in contravention to [the documentation](https://www.raspberrypi.org/documentation/hardware/raspberrypi/booteeprom.md)? I've read the following under `Automatic updates`:

>> The `rpi-eeprom-update` `systemd` service runs at startup and applies an update if a new image is available, automatically migrating the current bootloader configuration.

pelwell:
For "if a new image is available" read "if a suitable new recommended image is available".

An increase in power consumption at shutdown is worthy of an Issue/bug report: https://github.com/raspberrypi/rpi-eeprom/issues

lurch:
I think your question is answered by the comments in #321 ? "For the default bootloader release skip automatic updates unless the bootloader is older than a specified minimum version." (with that "specified minimum version" currently being 2020-09-03).

seamusdemora:
@lurch : Oh... I was hoping for clarification. :)

lurch:
What exactly do you want clarified? If you're using the `default` rpi-eeprom release stream, than a `sudo apt upgrade` will only automatically update you to 2020-09-03 (see https://github.com/raspberrypi/rpi-eeprom/blob/master/rpi-eeprom-update#L37 ). If you want a newer version, you can still run `sudo rpi-eeprom-update -a` which will get you the **very** newest `default` release, or you can switch to the `latest` or `beta` release streams.

seamusdemora:
I felt my question referenced a clear and un-ambiguous statement in the "official" documentation: 

>The `rpi-eeprom-update` `systemd` service runs at startup and applies an update if a new image is available, automatically migrating the current bootloader configuration. 

In my experience with English, that means that the firmware is updated if a newer version of the firmware is available. 

You qualified that: "specified minimum version", and after reading #321... 

If all of that is "clear", then perhaps I should try a do-over for 18 years of formal education :)

pelwell:
Which do you think needs changing - the behaviour or the documentation?

lurch:
Yes, the documentation-repo could possibly be clarified to be more "technically accurate", but I believe the assumption is that most users will be quite happy running with the `default` release-stream, and for those users there's no advantage in upgrading to a bootloader newer than 2020-09-03. If somebody really does want a newer bootloader version than that then they're probably best off switching to the `latest` release stream.

seamusdemora:
I'd vote for the [documentation](https://www.raspberrypi.org/documentation/hardware/raspberrypi/booteeprom.md). Perhaps that could be written to make it clear that only happens if one is following the `critical/default` stream in `/etc/default/rpi-eeprom-update` - or whatever the actual update rule is. As written, it sounds rather unequivocal: `... applies an update if a new image is available`

seamusdemora:
>  If somebody really does want a newer bootloader version than that then they're probably best off switching to the `latest` release stream.

Apparently that is what I was doing, and it has led to an unexpected result.

lurch:
If you're using the `latest` release stream, then a `sudo apt upgrade` **will** automatically update you to the newest `latest` EEPROM version.

timg236:
The Raspberry Pi Imager will always install the latest default release which works on all models and is inline with the latest release flashed at manufacture. There's no benefit in forcing an upgrade on older board revisions unless there's a bug fix.

However, if raspi-config, rpi-eeprom-update etc is used to change the bootloader configuration then the newest manufacturing release will be picked. This means you get important bug fixes e.g. network boot protocol fixes.

The latest feature (stable) release doesn't change. It's where the new features are.

Beta also doesn't change, but is likely to be dropped from APT. There's no benefit in delivering beta releases to everyone via APT which might cause a boot failure for the 0.1% of users who use beta releases.

lurch:
Just to clarify, I think when Tim says "The latest feature (stable) release doesn't change" I think he means that "The update logic used for `stable` eeprom updates isn't changed by #321 ". New eeprom images **will** still get published to the `latest` release stream in future (after they've gone through the `beta` stream first of course).

seamusdemora:
I've filed the bug report at #330. I'll update that with any new findings. 

Wrt how firmware releases are updated: At this point, I feel a bit like I've fallen in the rabbit's hole, but I'm sure it will make sense eventually. For now, I think I need to establish manual control over the firmware being used: I want to re-install my working version from the  `pieeprom-2020-06-15.bin`, and leave it there until this gets worked out. 

And thanks to all for your help - I do appreciate it.

lurch:
I think if you want to manually install an older eeprom release, you'll want to disable the systemd `rpi-eeprom-update` service to prevent `apt` automatically updating you to a newer version. See https://www.raspberrypi.org/documentation/hardware/raspberrypi/booteeprom.md

WRT your rabbit hole - yeah I think you might have confused yourself into thinking that it's more complicated than it actually is. The way I see it, the `default` release only updating to 2020-09-03 is a "little white lie" so that people using the `default` release stream don't end up with multiple bootloader updates that they don't actually need. But if anybody has manually switched to the `latest` or `beta` release streams, then `apt upgrade` always updates them to the newest version as they'd expect it to.

seamusdemora:
That *sounds like* there are different rules depending on which release stream (`beta`, `stable`/`latest` or `default`/`critical`) one follows? And I guess the names have been changed/augmented since I last looked at this (9 months ago) - I don't recall there being synonyms for the release streams.

Re the update service: Yes, I saw that when reading last night. It seems there are at least two methods:

1. `sudo systemctl mask rpi-eeprom-update` 

2. `FREEZE_VERSION` 

But to me, the `raspi-config` role still seems muddled in the documentation; i.e. it's not clear how these settings interact... does one override the other, etc. 

Re "confused myself": Perhaps - it wouldn't be the first time. But I'll add this comment: Making things "simple" is actually quite difficult. Personally, I find this quite removed from simple.

seamusdemora:
I have made some updates to the bug report at #330, including a discovery that using `dtoverlay=gpio-poweroff` in `/boot/config.txt` seems to play a role in the increased power consumption I am getting with `WAKE_ON_GPIO=0` and `POWER_OFF_ON_HALT=1` in the EEPROM configuration.

lurch:
> That sounds like there are different rules depending on which release stream (`beta`, `stable`/`latest` or `default`/`critical`) one follows?

Yes, was that not clear from the comments in #321 and the comments in this PR also?
"the `default` release only auto-updating to 2020-09-03 ... if anybody has manually switched to the `latest` or `beta` release streams, then apt upgrade always auto-updates them to the newest version as they'd expect it to."
If you really want the nitty-gritty: https://github.com/raspberrypi/rpi-eeprom/blob/master/rpi-eeprom-update#L610
(note that the `rpi-eeprom-update` [systemd service](https://github.com/raspberrypi/rpi-eeprom/blob/debian/buster/debian/rpi-eeprom.rpi-eeprom-update.service) runs the `rpi-eeprom-update` command with the `-s` option described here https://github.com/raspberrypi/rpi-eeprom/blob/master/rpi-eeprom-update#L414 )

> And I guess the names have been changed/augmented since I last looked at this

Yes, originally they were named `beta`, `stable` and `critical`; but as explained in the usage-text for rpi-eeprom-update, `latest` is now a symlink to `stable`, and `default` is now a symlink to `critical` https://github.com/raspberrypi/rpi-eeprom/blob/master/rpi-eeprom-update#L461
(the new names are more "user-friendly" but the old names are still needed for backwards-compatibility).

> sudo systemctl mask rpi-eeprom-update

This is effectively an OS-level setting, so that if you inserted this SD card into another RPi4, that wouldn't have its eeprom auto-updated either.

> FREEZE_VERSION

This is effectively an EEPROM-level setting, so that if you inserted a totally different RasPiOS SD card into this Pi, then your eeprom wouldn't get updated even if the `rpi-eeprom-update` systemd service was still running.

> what should the value of FIRMWARE_RELEASE_STATUS be if you don't wish to follow a stream at all?

If `FIRMWARE_RELEASE_STATUS` isn't set in `/etc/default/rpi-eeprom-update` then the update script will just fallback to the `default` release stream https://github.com/raspberrypi/rpi-eeprom/blob/master/rpi-eeprom-update#L27


BTW I'm sorry if my efforts to try and explain this have only confused you further :confused:

seamusdemora:
Not to worry - I am getting there... I don't have a great deal of time to invest in learning this - for me it is only the means to an end. Your efforts have helped, and I appreciate your time. 

And perhaps some good news as I've discovered that the dtoverlay for gpio-poweroff is causing my issue - I'll hope this conflict can be solved :)

PR #321, opened 2021-03-23
rpi-eeprom-update: Set a minimum version number for silent updates

timg236:
For the default bootloader release skip automatic updates unless the bootloader is older than a specified minimum version.

### DIFF
--- a/rpi-eeprom-update
+++ b/rpi-eeprom-update
@@ -34,6 +34,10 @@ VCMAILBOX=${VCMAILBOX:-/opt/vc/bin/vcmailbox}
 CM4_ENABLE_RPI_EEPROM_UPDATE=${CM4_ENABLE_RPI_EEPROM_UPDATE:-0}
 RPI_EEPROM_UPDATE_CONFIG_TOOL="${RPI_EEPROM_UPDATE_CONFIG_TOOL:-raspi-config}"
 
+# Automatic, critical updates are not applied unless the current bootloader version
+# is older than pieeprom-2020-09-03
+BOOTLOADER_AUTO_UPDATE_MIN_VERSION="${BOOTLOADER_AUTO_UPDATE_MIN_VERSION:-1599135103}"
+
 DT_BOOTLOADER_TS=${DT_BOOTLOADER_TS:-/proc/device-tree/chosen/bootloader/build-timestamp}
 
 EXIT_SUCCESS=0
@@ -301,9 +305,7 @@ checkDependencies() {
       BOARD_INFO="$(vcgencmd otp_dump | grep '30:' | sed 's/.*://')"
    fi
 
-   if [ $(((0x$BOARD_INFO >> 23) & 1)) -ne 0 ] && [ $(((0x$BOARD_INFO >> 12) & 15)) -eq 3 ]; then
-      echo "BCM2711 detected"
-   else
+   if [ $(((0x$BOARD_INFO >> 23) & 1)) -eq 0 ] || [ $(((0x$BOARD_INFO >> 12) & 15)) -ne 3 ]; then
       # Not a BCM2711, no EEPROMs to update.
       echo "This tool only works with a Raspberry Pi 4"
       exit ${EXIT_SUCCESS}
@@ -319,10 +321,8 @@ checkDependencies() {
    fi
 
    if [ ${BOARD_TYPE} -eq 17 ] && [ ${BOARD_REVISION} -lt 4 ]; then
-      echo "Dedicated VL805 EEPROM detected"
       HAVE_VL805_EEPROM=1
    else
-      echo "VL805 firmware in bootloader EEPROM"
       HAVE_VL805_EEPROM=0
    fi
 
@@ -409,7 +409,11 @@ Options:
    -l Returns the full path to the latest available EEPROM image file according
       to the FIRMWARE_RELEASE_STATUS and FIRMWARE_IMAGE_DIR settings.
    -m Write status information to the given file when run without -a or -f
-   -r Removes temporary EEPROM update files from the boot partition.
+   -r Removes temporary EEPROM update files from the boot partition. This also
+      reverts a pending update.
+   -s Skips silent, automatic upgrades for default releases if the current
+      bootloader release is newer than the the version specified by
+      BOOTLOADER_AUTO_UPDATE_MIN_VERSION ${BOOTLOADER_AUTO_UPDATE_MIN_VERSION}
    -u Install the specified VL805 (USB EEPROM) image file.
 
 Environment:
@@ -436,19 +440,18 @@ directory first. The binaries are then promoted to 'latest' and finally 'default
 so the 'default' binary is always the most tested release.
 
 default:
-The default bootloader image which is updated once new features in
-'latest' are stable or for critical hardware or security updates.
-
-Raspberry Pi OS automatically updates the bootloader when a newer 'default' image
-is available following an APT update to the rpi-eeprom package.
+The default bootloader image which supports all current models and hardware
+revisions.
+If a critical bug fix is required then the minimum default version number
+(BOOTLOADER_AUTO_UPDATE_MIN_VERSION) in the rpi-eeprom package is updated
+causing the bootloader to be automatically updated.
 
 latest:
 Contains the latest features which have undergone testing via the 'beta'
-release. The configuration parameters are stable.
-
-beta:
-Contains experimental features and bug fixes. Configuration parameters may
-change in subsequent 'beta' releases.
+release. Backwards compatiblity for configuration parameters is maintained
+once a feature is in the latest release directory.
+If the 'latest' release is selected then bootloader is automatically upgraded
+when the rpi-eeprom package is updated.
 
 As far as rpi-eeprom-update is concerned FIRMWARE_RELEASE_STATUS is just
 the subdirectory mapping under ${FIRMWARE_ROOT}. Therefore, custom release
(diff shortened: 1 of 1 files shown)


lurch:
Does this mean that some "critical" updates are less-critical than others? :wink:

rpi-eeprom-update
```diff
@@ -627,6 +632,20 @@ checkAndApply()
...
    fi
 
+   # If the shell is not interactive then skip silent upgrades for the default/critical release
+   # if the bootloader is already new enough. This avoids silent upgrades unless
+   # the minimum version is advanced but preserves the behaviour of "rpi-eeprom-update -a"
+   if ! tty -s; then
+      if [ -n "${BOOTLOADER_AUTO_UPDATE_MIN_VER}" ] && [ "${ACTION_UPDATE_BOOTLOADER}" = 1 ]; then
+         if [ "${FIRMWARE_RELEASE_STATUS}" = "critical" ] || [ "${FIRMWARE_RELEASE_STATUS}" = "default" ]; then
+            if [ "${BOOTLOADER_CURRENT_VERSION}" -ge "${BOOTLOADER_AUTO_UPDATE_MIN_VER}" ]; then
+               echo "Skip silent bootloader upgrade current: ${BOOTLOADER_CURRENT_VERSION} >= min ${BOOTLOADER_AUTO_UPDATE_MIN_VER}"
```
lurch, line 642: Super picky, but if there's a colon after `current` should there also be a colon after `min`?

timg236: the colon should be removed

timg236:
@waveform80 @vianpl @hiassoft 

I think this change should be compatible, can you see any problems with including it?

The purpose of this change is to make it easier to update the default release in Imager, raspi-config with any hardware support changes/fixes without doing as mass automatic update of current Pis. Instead of having product-specific releases or a big compatibility matrix, the default version would be updated but the minimum version is only advanced for important bug fixes that affect most/all models.

E.g. The 2021-03-18 should be the next default release but the minimum version will still be 2020-09-03. If the user doesn't touch the bootloader then there won't be an automatic upgrade. However, the following would upgrade as normal

Raspberry Pi Imager
raspi-config - boot-order / change version
"sudo rpi-eeprom-update -a" if run interactively

The interactive shell test is to avoid having another flag which means "yes, just update it"

rpi-eeprom-update
```diff
@@ -34,6 +34,10 @@ VCMAILBOX=${VCMAILBOX:-/opt/vc/bin/vcmailbox}
 CM4_ENABLE_RPI_EEPROM_UPDATE=${CM4_ENABLE_RPI_EEPROM_UPDATE:-0}
 RPI_EEPROM_UPDATE_CONFIG_TOOL="${RPI_EEPROM_UPDATE_CONFIG_TOOL:-raspi-config}"
 
+# Automatic, critical updates are not applied unless the current bootloader version
+# is older than pieeprom-2020-09-03
+BOOTLOADER_AUTO_UPDATE_MIN_VER="${BOOTLOADER_AUTO_UPDATE_MIN_VER:-1599135103}"
```
lurch, line 39: From a "separation of concerns" PoV, I can't decide whether or not it'd be better to have this in e.g. a config file (or symlink or something) in the `$FIRMWARE_IMAGE_DIR` directory, rather than having this `1599135103` value hardcoded in the `rpi-eeprom-update` script. Or if that'd be too fragile / error-prone.
 :confused:

timg236: It can already be overridden via /etc/default/rpi-eeprom-update or by setting the environment variable first. I don't think it merits a separate mechanism.

lurch: Ahh, I'd forgotten that `/etc/default/rpi-eeprom-update` gets sourced directly by the script, so can be used to override any of the variables' default values :+1:

timg236: Yep :)

HiassofT:
From a quick glance this change will break bootloader updates in LibreELEC unless we start manually adjusting the min version - which I'd like to avoid.

We run rpi-eeprom-update to check the current version against default (critical) release train and then present options to update VL805 and/or bootloader to the user.

If a user chooses to update then rpi-eeprom-update is called with `-A bootloader` and/or `-A vl805 via a non-interactive script / systemd service.

So if `-A bootloader` becomes a no-op then this would be not good...

HiassofT:
I think the behaviour of `-a` / `-A` should be kept as is.

If RPiOS prefers to have a different behaviour then it's best to make that opt-in and enable it only there - eg via another option or by not pre-setting `BOOTLOADER_AUTO_UPDATE_MIN_VER` in the script (that could be set eg via a config file on RPiOS).

I have to say I don't fully understand what problem this change tries to solve. You already have 3 release trains, splitting up the  default/critical into "really critical" and "not so critical" is quite confusing

timg236:
The latest default version needs to support every hardware revision, but silently upgrading bootloaders for older boards e.g. Pi 4B 1.1 that are only using SD card boot seems an unnecessary risk. The VL805 FW and R6 PMIC change was the reason for the upgrade to the 2020-09-03 release.

BOOTLOADER_AUTO_UPDATE_MIN_VER could be set to some large number to automatically upgrade from init scripts although the implementation could be cleaner for -a and -A.

Beta releases could be dropped from APT and replaced with the rpi-eeprom-recovery-beta.zip in Github for the latest beta. That just gives default for automatic updates and latest for people who want new features.

HiassofT:
I think a separate option for this new behaviour would make most sense. Then you could also drop the interactive tty check - automagically switching operation modes like that is rather unexpected and once you'd add automated CI tests you'd hate yourself for doing such things :-)

timg236:
I think it would be to add a '-s' flag to indicate that this is a silent update so no change to "rpi-eeprom-update -a"

'rpi-eeprom-config -s' will exit with 0 if the bootloader is "new enough"
'rpi-eeprom-config -s 'a' will exit with 0 and do nothing if the bootloader is "new enough"

The minimum version can be included in the machine output

The Raspberry Pi OS APT package will specify have the following in the rpi-eeprom-update.service 
ExecStart=/usr/bin/rpi-eeprom-update -s -a

lurch:
A 'silent' flag typically tells a program to run normally, but don't print anything to stdout, so probably isn't suitable here?
I guess you could keep it as `-s` but label it "Skip auto-update if EEPROM already new enough" ? :wink:

timg236:
I'll do some more testing tomorrow, hopefully, this is a better compromise for system to automatically 'do the right thing' whilst preserving backwards compatibility.

rpi-eeprom-update
```diff
@@ -707,17 +731,18 @@ BOOTLOADER_LATEST=${bootloader_new}
...
 VL805_LATEST="${vl805_new}"
 VL805_EEPROM="${vl805_eeprom}"
+BOOTLOADER_AUTO_UPDATE_MIN_VERSION=${min_ver}
 EOF
    else
       [ "${HAVE_VL805_EEPROM}" = "0" ] && vl805_eeprom="false" || vl805_eeprom="true"
       cat > "${MACHINE_OUTPUT}" <<EOF
 {
   "EXITCODE": "${exit_code}",
-  "BOOTLOADER_CURRENT": ${bootloader_cur},
```
lurch, line 716: Seems strange to make this change in the `else` branch but not the `if` branch? (was this removed accidentally?)

rpi-eeprom-update
```diff
@@ -707,17 +731,18 @@ BOOTLOADER_LATEST=${bootloader_new}
...
   "BOOTLOADER_LATEST": ${bootloader_new},
   "VL805_CURRENT": "${vl805_cur}",
   "VL805_LATEST": "${vl805_new}",
   "VL805_EEPROM": ${vl805_eeprom}
```
lurch, line 744: I think you need a comma at the end of this line to keep the JSON-formatting.

rpi-eeprom-update
timg236: Yes, clearly it was too late in the evening to be writing scripts

rpi-eeprom-update
```diff
@@ -597,6 +606,19 @@ lookupVersionInfo()
...
 
+   # If the '-s' flag for silent updates is specified then only update the
+   # bootloader if the current version is older than the minimum version.
+   if [ ${SILENT_UPDATE} = 1 ] && [ -n "${BOOTLOADER_AUTO_UPDATE_MIN_VERSION}" ] && [ "${ACTION_UPDATE_BOOTLOADER}" = 1 ]; then
+      if [ "${FIRMWARE_RELEASE_STATUS}" = "critical" ] || [ "${FIRMWARE_RELEASE_STATUS}" = "default" ]; then
+         if [ "${BOOTLOADER_CURRENT_VERSION}" -ge "${BOOTLOADER_AUTO_UPDATE_MIN_VERSION}" ]; then
+            echo "Skip automatic bootloader upgrade. current ${BOOTLOADER_CURRENT_VERSION} >= min ${BOOTLOADER_AUTO_UPDATE_MIN_VERSION}"
+
+            # Clear the update requried flag
+            ACTION_UPDATE_BOOTLOADER=0
```
lurch, line 617: Should this also set `BOOTLOADER_UPDATE_IMAGE=""` like you do inside `checkAndApply` (and on line 606 above) ?

timg236:
Cleared the variable as suggested for clarity, although it's not looked at. I've also tidied up the human-readable version output

rpi-eeprom-update
```diff
@@ -436,19 +442,20 @@ directory first. The binaries are then promoted to 'latest' and finally 'default
...
-'latest' are stable or for critical hardware or security updates.
-
-Raspberry Pi OS automatically updates the bootloader when a newer 'default' image
-is available following an APT update to the rpi-eeprom package.
+The default bootloader image which supports all current models and hardware
+revisions.
+If a critical bug fix is required then the minimum default version number
+BOOTLOADER_AUTO_UPDATE_MIN_VERSION in the rpi-eeprom package is updated
+causing the bootloader to be automatically updated. Otherwise, if the 'default'
+bootloader release is selected then the bootloader is only updated if the
```
lurch, line 450: "the bootloader is only updated if..." - is this text still relevant now that you've added the `-s` option?

PR #327, opened 2021-04-20
rpi-eeprom-update: clarify docs

andrum99:
Mention that if -d is not specified, the current bootloader config will be retained.

### DIFF
--- a/rpi-eeprom-update
+++ b/rpi-eeprom-update
@@ -400,6 +400,7 @@ Options:
       flag use the config in that file. This option only applies when a
       bootloader EEPROM update is needed; if the bootloader EEPROM is up-to-date
       then its config will not be changed.
+      Without this switch, the current bootloader config will be retained.
    -f Install the given file instead of the latest applicable update
       Ignores the FREEZE_VERSION flag in bootloader and is intended for manual
       firmware updates.


pelwell:
This patch is a bit belt-and-braces. If the corollary isn't obvious enough in itself, just a few lines up it says:
> Unless the -d flag is specified, the current bootloader configuration is retained.

rpi-eeprom-update
```diff
@@ -400,6 +400,7 @@ Options:
       flag use the config in that file. This option only applies when a
       bootloader EEPROM update is needed; if the bootloader EEPROM is up-to-date
       then its config will not be changed.
+      Without this switch, the current bootloader config will be retained.
```
lurch, line 403: See line 392? :wink:

timg236: I think the preservation of the EEPROM config is already specified.
The "Use the default bootloader config" text could be removed because there is no default bootloader-config, just whatever the eeprom image contains.
aka -d - don't migrate EEPROM config

andrum99:
> This patch is a bit belt-and-braces. If the corollary isn't obvious enough in itself, just a few lines up it says:
> 
> > Unless the -d flag is specified, the current bootloader configuration is retained.

Apologies folks - I should have read this more carefully. I hadn't spotted that this was already mentioned. Closing.

PR #337, opened 2021-05-19
Pieeprom 2021 05 19

timg236:
@peterharperuk 

I've tested the beta bootloader, please could you have a quick check of the script updates?


peterharperuk:
I spent some time playing with this. Change seems good to me

PR #343, opened 2021-06-10
Add a test to cause the script to exit cleanly on non-Pi hardware

waveform80:
At present, the update script exits cleanly in the case it finds itself run on Pi hardware prior to the 4/CM4/400, but if installed on non-Pi hardware exits with various slightly cryptic errors (missing `vcgencmd` and some subsequent math errors from trying to bit-shift the empty `BOARD_INFO` variable).

This patch adds a similar exit to the non-Pi 4 case, with a slightly more useful message. It may seem rather academic but as rpi-eeprom is now in the Ubuntu armhf and arm64 archives it can theoretically (although not sensibly :) be installed on something like an AWS arm64 container.

It's arguable whether it should be a "clean" exit in this scenario. However, consider the example of a non-Pi system which has a "Revision" entry in `/proc/cpuinfo` (although obviously not a Pi revision). In this case, this script will currently exit cleanly stating "This tool only works with a Raspberry Pi 4". Likewise for a non-Pi board with a `linux,revision` node in its device-tree. Hence my choice to stick with the pattern and echo a message followed by a clean exit, but obviously I'm happy to revise this if wanted?

### DIFF
--- a/rpi-eeprom-update
+++ b/rpi-eeprom-update
@@ -315,8 +315,11 @@ checkDependencies() {
       BOARD_INFO="$(od -v -An -t x1 /sys/firmware/devicetree/base/system/linux,revision | tr -d ' \n')"
    elif grep -q Revision /proc/cpuinfo; then
       BOARD_INFO="$(sed -n '/^Revision/s/^.*: \(.*\)/\1/p' < /proc/cpuinfo)"
-   else
+   elif command -v vcgencmd > /dev/null; then
       BOARD_INFO="$(vcgencmd otp_dump | grep '30:' | sed 's/.*://')"
+   else
+      echo "No Raspberry Pi board info found"
+      exit ${EXIT_SUCCESS}
    fi
 
    if [ $(((0x$BOARD_INFO >> 23) & 1)) -eq 0 ] || [ $(((0x$BOARD_INFO >> 12) & 15)) -ne 3 ]; then


berlin4apk:
Why not exit with 128?
"the value 128 is used to indicate failure to execute another program in a subprocess. This convention is not universally obeyed, but it is a good idea to follow it in your programs." https://www.gnu.org/software/libc/manual/html_node/Exit-Status.html

berlin4apk:
Can the PR together with this maybe leed to a logic error?
https://github.com/raspberrypi/rpi-eeprom/blob/71ef83abbcfc6cd43a8f5cb7570f341e55a49e9b/rpi-eeprom-update#L354

waveform80:
> Why not exit with 128?
> "the value 128 is used to indicate failure to execute another program in a subprocess. This convention is not universally obeyed, but it is a good idea to follow it in your programs." https://www.gnu.org/software/libc/manual/html_node/Exit-Status.html

Only because it's not in keeping with the existing behavior (see examples in the description of a system with a "Revision" field in `/proc/cpuinfo` or a `linux,revision` node in the device-tree). I'm open to the idea of using such an exit code, but if that's the case I think the other circumstances should probably be updated to exit with such a code.

There's also the question of whether the error requires meaningful action on the part of the system administrator. For instance, it's perfectly reasonable that the script might wind up being run on a Pi 3 if an SD card is moved from a Pi 4, in which case a clean exit makes sense (yes, the script is technically running on the "wrong" system, but the card could well be moved back to the Pi 4). Likewise, if there's a reasonable possibility of an image which could operate on both a Pi and a non-Pi ARM system, then there's a similar argument that a clean exit is reasonable in this case.

Overall at the moment it's this latter case which sways me towards a clean exit code and message, but I'm happy to be persuaded otherwise! :)

waveform80:
> Can the PR together with this maybe leed to a logic error?
> 
> https://github.com/raspberrypi/rpi-eeprom/blob/71ef83abbcfc6cd43a8f5cb7570f341e55a49e9b/rpi-eeprom-update#L353-L354

Interesting point; I'd argue that indicates that the script already has a slightly odd flow in that it attempts to utilize `vcgencmd` at the top of `checkDependencies` and *then* checks whether `vcgencmd` is installed later. However, it is actually a valid check given that the path taken *could* be:

1. Line 314, test for `/sys/firmware/devicetree/base/system/linux,revision` fails (old firmware)
2. Line 316, test for "Revision" in `/proc/cpuinfo` succeeds (kernel has Pi patches)
3. Line 325, test for Pi 4 compatible code in `BOARD_INFO` succeeds (we're running on a Pi 4)
4. Line 352, test for bootloader timestamp in device-tree fails (old firmware)
5. Line 353, test for `vcgencmd` fails (because libraspberrypi-bin not installed)

lurch:
There's also the fact that (at least on Raspberry Pi OS) `rpi-eeprom-update` is run by one of the startup-scripts, so doing a "clean exit" probably makes the management of that simpler too :slightly_smiling_face:

lurch:
> the script already has a slightly odd flow

I think that just reflects the fact that different parts of the script were edited at different times, with various attempts at improving "cross OS compatibility" with all the various OSes that want to make use of this script :wink:

timg236:
Looks ok to me. I'll give it a quick test before merging tomorrow

timg236:
Looks good. Thanks

PR #362, opened 2021-09-22
recovery.bin: Fix recovery.bin loading on larger FAT partitions - BETA

timg236:


timg236:
@peterharperuk - Please can you check this fixes the OP's issue. If we are happy with this then we should update recovery.bin in APT + stable/default.

peterharperuk:
I tested it on my device and it fixes the problem. I'll ask the OP if they can test.

PR #363, opened 2021-09-23
Promote 22/09/2021 beta recovery.bin to stable

timg236:
@peterharperuk @XECDesign 

Let's bump this from beta to stable. This means that people can easily compare the previous and the new release using raspi-config default / latest without having to switch to beta releases.


XECDesign:
Ah sorry, I've just pushed the previous version without this PR. I can update again when this is merged.

timg236:
@XECDesign No problem. It's useful to have the different versions in-case people want to bisect changes

PR #369, opened 2021-10-04
pieeprom-2021-10-04: Add support for GPT FAT16 and increase USB timeouts - BETA

timg236:
* Update the FAT detection to support FAT16 for EFI/ESD paritions with
  GPT instead of assuming FAT32. The latest firmware is also required
  for a similar update.
* Increase the timeouts for MSD SCSI commands to reduce the risk of
  timeouts when probing the capacity of slow to start devices
  e.g. USB RAID with spinning disks.


firmware/release-notes.md
```diff
@@ -3,6 +3,14 @@
 USB MSD boot also requires the firmware from Raspberry Pi OS 2020-08-20 or newer.
 https://www.raspberrypi.org/documentation/hardware/raspberrypi/bcm2711_bootloader_config.md
 
+## 2021-10-04 - Add support for GPT FAT16 and increase USB timeouts - BETA
+  * Update the FAT detection to support FAT16 for EFI/ESD paritions with
```
lurch, line 7: ESD -> ESP ? https://en.wikipedia.org/wiki/EFI_system_partition

PR #371, opened 2021-10-06
Fix URLs in README

peterharperuk:
These got broken by the recent doc update and the redirects to
raspberrypi.com don't seem to be correct


peterharperuk:
@timg236

aallan (approved):
LGTM

PR #380, opened 2021-11-22
2021-11-22: Fix for Sabrent rocket Nano NVMe reboot issue - BETA

peterharperuk:
* Fixes issue with Sabrent rocket Nano NVMe disk after a reboot.
  Run pcie initialisation again if there's an error.


peterharperuk:
@timg236

PR #384, opened 2021-12-09
Update default recovery.bin to the latest stable release

timg236:
* Promote the recovery.bin from stable to default. This avoids an
  issue where recovery.bin fails to load on large FAT32 boot partions with
  32K clusters.


timg236:
@peterharperuk

peterharperuk:
Seems sensible

PR #385, opened 2021-12-09
2021-12-09: Fix MTB detection for factory test - BETA

peterharperuk:
* Just fixes a regression with MTB detection affecting factory testing


peterharperuk:
@timg236

PR #386, opened 2021-12-10
2021-12-02: Fix MTB detection for factory test - BETA

peterharperuk:
* Just fixes a regression with MTB detection affecting factory testing

Note: This is just a rename of 2021-12-09 which had the wrong date


peterharperuk:
@timg236

timg236:
LGTM

PR #399, opened 2022-01-24
Add note about GPU memory in latest beta

timg236:


timg236:
@peterharperuk

lurch:
Should we add this to the documentation too? https://www.raspberrypi.com/documentation/computers/config_txt.html#memory-options

Or is this such an unlikely use-case that it's not worth mentioning? ping @aallan

PR #400, opened 2022-01-25
pieeprom-2022-01-25: Create stable release from pieeprom-2022-01-20

timg236:
@peterharperuk


peterharperuk:
Why is there a new beta as well?

timg236:
> Why is there a new beta as well?

It had to be rebuilt for secure-boot, I think it's easier to understand if the stable releases also appear in beta.

peterharperuk:
ok - understood. Looks good then

PR #404, opened 2022-02-03
make-release: Use rpi-eeprom-digest to generate update timestamps

timg236:
Previously, the .sig would have an update timestamp of zero. Switching
to rpi-eeprom-digest is preferable because the update timestamp will
be much closer to when the image was installed.

N.B. Update-timestamp is used to track configuration changes and is
independent of the build-timestamps which are the version of the
bootloader executable.


timg236:
@peterharperuk

peterharperuk:
Looks good

PR #407, opened 2022-02-09
pieeprom-2022-02-08: Fix secure-boot boot failure - STABLE

timg236:
Fix boot failure regression on boards which had the OTP secure boot bits set.


timg236:
@peterharperuk

peterharperuk:
Yes - seems goof

PR #411, opened 2022-02-17
2022-02-16: Network Install updated - BETA

peterharperuk:
Net install changes.
* Got rid of confirmation step that required you to press <Space> to
  initiate net install. Now just long press <Shift>
* Updated the screen text to make it more obvious the device is still
  trying boot when the net install is showing.
* Fixed a DHCP net install bug which caused us to lose the
  gateway address.
* Fixed a bug with the uIP timers which could cause net install to
  always fail.
* Implemented resume and retry on download failure.

Other interesting changes.
* Allow smaller MSD discovery timeouts to be specified.
* Some tweaks and fixes to IPV6 netboot.


peterharperuk:
@timg236

lurch:
Ahhh, so they were pressing Shift, and then inserting a bootable SD-card before pressing space, rather than after?

peterharperuk:
yes, or they saw the net install screen appear and thought it indicated that their device was not going to boot.

PR #413, opened 2022-02-20
Improve BOOTFS detection

JustArchi:
This PR adds following enhancements:
- Adds probing for common, alternative `BOOTFS` mount point, `/boot/firmware` which can be found for example in Debian images for Raspberry Pi. This happens only when `BOOTFS` was not specified by the caller, and default `/boot` location doesn't contain `.elf` files.
- Improves `BOOTFS` empty warning by limiting depth to `1`. This is important, as files deeper should not be considered, even if they have appropriate extension. In my case, the warning was not triggered despite no `.elf` files in my `/boot` - this is because `find` picked files in `/boot/firmware` instead and was happy with that.
- When removing/reverting update, prints detected `BOOTFS` in similar manner as install does.

Context: I was attempting to update my RPI's bootloader today on unofficial Debian image, which mounts RPI's boot partition with `.elf` blobs under `/boot/firmware`. This obviously has failed, as `rpi-eeprom-update` assumed that the partition is available under `/boot` instead. While currently supported `BOOTFS=/boot/firmware ./rpi-eeprom-update` way of launching the script is sufficient for my use case, I came to conclusion that it shouldn't be too much of a burden to probe that location in addition to falling back to `/boot` as right now. While doing that I also found two other misc things which I fixed by the way while coding the main feature.

I coded the probing feature in a way it triggers if, and only if, `BOOTFS` was left unspecified, which still allows overriding the detection to whatever folder caller finds appropriate. The logic also triggers only if default `/boot` has no `.elf` files inside - so we can be sure that whatever we do, `/boot` is not the correct place regardless. I also made it very easy to extend for more paths in the future if need arises - as long as they don't contain spaces, we're fine maintaining POSIX `sh` compatibility.

Naturally, after those changes, `rpi-eeprom-update -b` now prints `/boot/firmware` on my machine. It also doesn't cause any regression in official images, and I don't believe somebody could accidentally run into alternative path - he'd need to have no `.elf` files in `/boot` and `.elf` files in `/boot/firmware`, you can't do it accidentally, and even if you did, then previously used `/boot` is wrong path regardless in this case.

Thank you in advance for considering those enhancements.

### DIFF
--- a/rpi-eeprom-update
+++ b/rpi-eeprom-update
@@ -29,7 +29,10 @@ FIRMWARE_IMAGE_DIR=${FIRMWARE_IMAGE_DIR:-${FIRMWARE_ROOT}/${FIRMWARE_RELEASE_STA
 FIRMWARE_BACKUP_DIR=${FIRMWARE_BACKUP_DIR:-/var/lib/raspberrypi/bootloader/backup}
 ENABLE_VL805_UPDATES=${ENABLE_VL805_UPDATES:-1}
 RECOVERY_BIN=${RECOVERY_BIN:-${FIRMWARE_ROOT}/${FIRMWARE_RELEASE_STATUS}/recovery.bin}
+
 BOOTFS=${BOOTFS:-/boot}
+BOOTFS_ALTERNATIVE_PATHS="/boot/firmware"
+
 CM4_ENABLE_RPI_EEPROM_UPDATE=${CM4_ENABLE_RPI_EEPROM_UPDATE:-0}
 RPI_EEPROM_UPDATE_CONFIG_TOOL="${RPI_EEPROM_UPDATE_CONFIG_TOOL:-raspi-config}"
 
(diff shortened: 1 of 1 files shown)


rpi-eeprom-update
```diff
@@ -544,10 +549,22 @@ findBootFS()
       BOOTFS="${TMP_BOOTFS_MNT}"
    fi
 
+   # Some distros mount boot partition in alternative locations, e.g. Debian
+   # We can aid them by probing the most common alternative paths
+   if [ -n "${BOOTFS_ALTERNATIVE_PATHS}" ]; then
```
lurch, line 554: Should this be in the `else` clause of the previous `if` block?

Also, if `$BOOTFS` (i.e. `/boot`) _does_ contain an `.elf` file, perhaps it should skip checking `$BOOTFS_ALTERNATIVE_PATHS`? :man_shrugging:

rpi-eeprom-update
```diff
@@ -29,7 +29,12 @@ FIRMWARE_IMAGE_DIR=${FIRMWARE_IMAGE_DIR:-${FIRMWARE_ROOT}/${FIRMWARE_RELEASE_STA
 FIRMWARE_BACKUP_DIR=${FIRMWARE_BACKUP_DIR:-/var/lib/raspberrypi/bootloader/backup}
 ENABLE_VL805_UPDATES=${ENABLE_VL805_UPDATES:-1}
 RECOVERY_BIN=${RECOVERY_BIN:-${FIRMWARE_ROOT}/${FIRMWARE_RELEASE_STATUS}/recovery.bin}
-BOOTFS=${BOOTFS:-/boot}
+
+if [ -z "${BOOTFS}" ]; then
```
timg236, line 33: The -z BOOTFS line is unnecessary  the line above does that

timg236:
Why not just specify BOOTFS=/boot/firmware in /etc/default/rpi-eeprom-update?

rpi-eeprom-update
JustArchi: I've addressed both, thanks!

JustArchi:
> Why not just specify BOOTFS=/boot/firmware in /etc/default/rpi-eeprom-update?

No good reason, same as calling `BOOTFS=/boot/firmware ./rpi-eeprom-update` - both will work and there is nothing wrong in them, in fact I appreciate those ways exist as it's exactly how I used it myself. In this PR I'm merely making it easier and config-free for other distros to use `rpi-eeprom-update` without forcing them to create standalone config files. Whether that is appropriate or not, I leave up to you. Personally I don't consider it a huge maintenance burden, and as it's coded right now, the alternative logic will happen if, and only if, user does not specify `BOOTFS` through config or env variable, and his default `/boot` has no `.elf` files. Regardless of the situation, putting stuff in `/boot` is wrong in this case, so alternative detection can help probing for the most common locations, I see no harm done in that.

timg236:
The patch seems reasonable, but I don't want to end up having a list of plausible boot locations for different operating systems and then debate ordering etc. The script and default config are for Raspberry Pi OS only and the customization point would be the defaults file or a fork if really necessary.

I'll leave it open for a couple of days to see if there are comments from other OS distro maintainers.

PR #415, opened 2022-02-28
2022-02-28: More net Install changes - BETA

peterharperuk:
Net install changes.
* Net install is initiated on boot if shift is pressed.
* New HTTP boot order (7) and configuration parameters,
  HTTP_HOST, HTTP_PATH, HTTP_PORT to set url

Other interesting changes.
* Increase the max ramdisk size to 128MB
* Increase timeout of early SD/EMMC commands to 100ms


lurch:
> configuration parameters, HTTP_HOST, HTTP_PATH, HTTP_PORT to set url

Is that the URL of the embedded-RPi-Imager, or the URL of the JSON file that RPi-Imager uses?

peterharperuk:
It's the URL of the embedded image - which can be the rpi imager image or anything else in theory. I've posted a link to some draft documentation here https://forums.raspberrypi.com/viewtopic.php?t=329324&start=150#p1979161

lurch:
I wonder if there's a use-case for **not** changing the `HTTP_HOST` setting, but instead having some other option which effectively replicates the `--repo` option of RPi Imager? https://github.com/raspberrypi/rpi-imager#custom-repository

peterharperuk:
You can already add the imager repo url to the eeprom config? Is that what you meant?

IMAGER_REPO_URL=https://www.yourserver.acme/oslist.json

PR #425, opened 2022-04-26
pieeprom-2022-04-26: Network install / secure-boot STABLE/BETA

timg236:
Candidate for the next default release.


timg236:
@peterharperuk FYI this release enables network-install / HTTP download to be used with secure-boot

lurch:
How is this both "beta" **and** "stable" ? :confused:

peterharperuk:
@lurch Network install is in both beta and stable, so it's just the signing key that's changed.
@timg236 Do you want me to test this on a "secure" pi?

lurch:
Ahhh, so the `pieeprom-2022-04-26.bin` in the beta directory is actually different to the `pieeprom-2022-04-26.bin` in the stable directory? (If so, I think that might be the first time that this has happened, which is why I was confused)

timg236:
> How is this both "beta" **and** "stable" ? 😕

Same binary

timg236:
> @lurch Network install is in both beta and stable, so it's just the signing key that's changed. @timg236 Do you want me to test this on a "secure" pi?

Yes, please. I have a pending commit which updates all the binaries in usbboot

PR #428, opened 2022-05-11
rpi-eeprom-update: Avoid using recovery.bin if the current bootloader supports self-update

timg236:
Self-update is preferred to using recovery.bin because it avoids modifiy the
boot partition in order to rename recovery.bin after use. Since the 2711 ROM
does not support network or USB MSD loading of recovery.bin self-update has to
be used with other boot modes anyway.

If RPI_EEPROM_SELF_UPDATE=1 then avoid installing recovery.bin so long as the
current bootloader version supports self-update from SD/MMC and that doesn't
look as though SELF_UPDATE has been disable in the EEPROM config.

RPI_EEPROM_SELF_UPDATE is currently set to 0 by default because this is is an experimental change.

### DIFF
--- a/rpi-eeprom-update
+++ b/rpi-eeprom-update
@@ -33,6 +33,17 @@ BOOTFS=${BOOTFS:-/boot}
 CM4_ENABLE_RPI_EEPROM_UPDATE=${CM4_ENABLE_RPI_EEPROM_UPDATE:-0}
 RPI_EEPROM_UPDATE_CONFIG_TOOL="${RPI_EEPROM_UPDATE_CONFIG_TOOL:-raspi-config}"
 
+# Self-update is preferred to using recovery.bin because it avoids modifiy the
+# boot partition in order to rename recovery.bin after use. Since the 2711 ROM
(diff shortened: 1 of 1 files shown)


rpi-eeprom-update
```diff
@@ -33,6 +33,17 @@ BOOTFS=${BOOTFS:-/boot}
...
 
+# Self-update is preferred to using recovery.bin because it avoids modifiy the
+# boot partition in order to rename recovery.bin after use. Since the 2711 ROM
+# does not support network or USB MSD loading of recovery.bin self-update has to
+# be used with other boot modes anyway.
+
+# If RPI_EEPROM_SELF_UPDATE=1 then avoid installing recovery.bin so long as the
+# current bootloader version supports self-update from SD/MMC and that doesn't
+# look as though SELF_UPDATE has been disable in the EEPROM config or config.txt.
+RPI_EEPROM_SELF_UPDATE="${RPI_EEPROM_SELF_UPDATE:-0}"
```
peterharperuk, line 44: Should this be defaulting to 1?

PR #429, opened 2022-05-20
pieeprom-2022-05-20.bin Reduce boot-time when network install is disabled. - BETA

timg236:
* Reduce boot-time when network install is disabled NET_INSTALL_ENABLED=0.
* Switch to the newer SDIO HC and increase SPI clock speed.


timg236:
@peterharperuk

peterharperuk:
Looks ok

andrum99:
> * Switch to the newer SDIO HC ...

Presumably this means that the EEPROM boot firmware is now using the new DDR-capable SDIO block to talk to the SD card during SD boot?

timg236:
> > * Switch to the newer SDIO HC ...
> 
> Presumably this means that the EEPROM boot firmware is now using the new DDR-capable SDIO block to talk to the SD card during SD boot?

The bootloader will still operate at HS mode, it just removes some quirks of the legacy SDIO controller

symbios24:
Hello,What is the overal speed improvement?

timg236:
> Hello,What is the overal speed improvement in the memory card?

Not much, it's running at HS as before. Just a 50ms saving from a defunct HC reset. It does allow some legacy quirks to be removed so it's just a note for reference e.g. in the unlikely event it changes behavior on some cards.
The changes to skip USB init if network-install is disabled was more significant.

symbios24:
Thanks, i assume its not possible for the sd card controller to work as sdf104 ?

lurch:
> Thanks, i assume its not possible for the sd card controller to work as sdr104 ?

This repo _only_ deals with the SD card controller modes used by the eeprom bootloader (to load `start4.elf` and `kernel.img` etc.). The [Linux kernel](https://github.com/raspberrypi/linux) determines what SD card controller modes are used by Linux itself.

> Do i have to disable the netboot from inside the epprom or its disabled by default?

See https://www.raspberrypi.com/documentation/computers/raspberry-pi.html#raspberry-pi-4-bootloader-configuration

PR #435, opened 2022-07-15
2022-07-14 - Add pieeprom-2022-07-14 - BETA

timg236:
Enable long-filenames & sub-directories for start_file & fixup_file.
Use Unix path separators with a maximum path of 255 characters.
Relative paths (. or ..) are not supported.


timg236:
@peterharperuk

PR #446, opened 2022-10-18
2022-10-18 - Tryboot enhancements for A/B partition booting - BETA

timg236:
* Add support for a [tryboot] conditional statement in config files.
* Load config.txt instead of tryboot.txt if tryboot_a_b=1 in autoboot.txt
* Fix failover to partition 1  if the `boot_partition` points to non-bootable partition.
* Enable `autoboot.txt` in secure-boot mode.


timg236:
@peterharperuk

PR #448, opened 2022-11-02
pieeprom-2022-11-02: Add option to use Customer OTP for MAC address -…

timg236:
… BETA

Add a new EEPROM property that allows the Ethernet MAC address programmed during manufacture to be overridden a value in the Customer OTP register.

MAC_ADDRESS_OTP=A,B
where A and B are the customer row numbers (0..7)


timg236:
@peterharperuk @jamesh65

peterharperuk:
Looks good.

PR #455, opened 2022-11-28
rpi-eeprom-config: Improve robustness of padding code

timg236:
Make it obvious that the modifiable files are 4K aligned regions in the EEPROM and add support for files with different padding (for future use).


timg236:
@peterharperuk Some preliminary changes that fell out of the custom CA cert investigations.

rpi-eeprom-config
```diff
@@ -39,6 +37,11 @@ FILE_HDR_LEN = 20
 FILENAME_LEN = 12
 TEMP_DIR = None
 
+# Modifaiable files are stored in a single 4K erasable sector.
```
peterharperuk, line 40: Modifaiable => Modifiable

peterharperuk:
Can't spot any issues

rpi-eeprom-config
timg236: Fixed

PR #457, opened 2022-12-07
pieeprom-2022-12-07

timg236:
Fix SD voltage voltage reset on Pi4 R1.1 - the regression occurred in the stable release  pieeprom-2022-07-19 but wasn't spotted until now.


timg236:
@peterharperuk

peterharperuk:
Looks ok

PR #461, opened 2023-01-10
"[rpi-eeprom-update] Fix rpi-eeprom-config path"

goldyfruit:
For some reason, when running `sudo rpi-eeprom-update -a`, the newer available firmware is never applied even if the command doesn't return any error.

When running `sudo rpi-eeprom-config -a` in debug, `rpi-eeprom-config` command appears to not be found because the script is looking into `/home/pi` instead of `/usr/bin`.

```
+ /home/pi/rpi-eeprom-config --out /tmp/tmp.mBRSuU7naU --config /tmp/tmp.MXSyIs3I0u /lib/firmware/raspberrypi/bootloader/beta/pieeprom-2023-01-04.bin
/usr/bin/rpi-eeprom-update: line 178: /home/pi/rpi-eeprom-config: No such file or directory
```

### DIFF
--- a/rpi-eeprom-update
+++ b/rpi-eeprom-update
@@ -5,6 +5,7 @@
 set -e
 
 script_dir=$(cd "$(dirname "$0")" && pwd)
+rpi_eeprom_config_dir="/usr/bin"
 
 if [ -f /etc/default/rpi-eeprom-update ]; then
    . /etc/default/rpi-eeprom-update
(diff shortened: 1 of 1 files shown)


timg236:
Hardcoding /usr/bin here is the wrong approach. Maybe rpi-eeprom-config could just rely upon the search path but I don't see a strong reason to change this to support running the scripts from different directories

pelwell:
Agreed - it's perfectly reasonable to expect rpi-eeprom-config and rpi-eeprom-update to be in the same directory, whatever that may be.

PR #464, opened 2023-01-13
pieeprom-2023-01-11: Promote previous BETA release to STABLE

timg236:
This commit updates the VL805 firmware to vl805-000138c0.bin 
See https://github.com/raspberrypi/linux/pull/5262


timg236:
@peterharperuk Seems good so let's update the APT stable package next week.

peterharperuk:
ok, makes sense

PR #482, opened 2023-08-01
firmware: Remove the beta release folder

timg236:
Remove the beta release folder and replace with a symlink to the 'latest' release.

Including beta firmware in the APT release causes unecessary churn and occasionally causes an issue where a user might end up on an old and unsupported beta release.

Instead, beta releases are will be feature specific and installed manually via 'rpi-eeprom-config --edit beta-firmware.bin'

i.e. beta releases will be out-of-package custom firmware releases downloaded from github. The functionality will be merged into a new latest/stable release or abandoned.

N.B. The old beta releases have been archived to 'old' but are no longer included in APT.


timg236:
@peterharperuk @XECDesign

lurch:
I just checked (out of curiosity), and `raspi-config` uses:
```
/lib/firmware/raspberrypi/bootloader/stable/
/lib/firmware/raspberrypi/bootloader/critical/
```
:slightly_smiling_face:

timg236:
> I just checked (out of curiosity), and `raspi-config` uses:
> 
> ```
> /lib/firmware/raspberrypi/bootloader/stable/
> /lib/firmware/raspberrypi/bootloader/critical/
> ```
> 
> 🙂

I've created a PR to update raspi-config to use the desired names. I've also updated it to prefer rpi-eeprom-config --apply rather than the older/low-level commands

firmware/stable/beta
```diff
@@ -0,0 +1 @@
+beta
```
XECDesign, line 1: Is this meant to be here?

timg236:
Removed the errant beta->beta link

PR #485, opened 2023-09-28
Add support for Pi5

timg236:


timg236:
@XECDesign @peterharperuk

timg236:
@popcornmix

HiassofT:
Thanks a lot, Tim, testing on RPi4 and RPi5 looks fine here (on LibreELEC).

Only minor nit: 2711 firmware is in the "legacy style" critical/stable directories whereas 2712 one in the "new style" default/lastest directories.

This makes packaging a bit trickier for us as we have to distinguish between 2711 and 2712, look at different directories and create different symlinks.

Not a huge deal, altough if it doesn't break other stuff it might be a good time to move the 2711 directory structure to the new style as well - due to the firmware->firmware-2711 change people need to adjust their packaging scripts anyways

timg236:
Yeah, it might be worth fixing up the 2711 directory to have the same structure as the new-style. Will check with others tomorrow.

timg236:
I've updated the 2711 release names as discussed

HiassofT:
Thanks a lot, this looks fine now, and tests on RPi4 and RPi5 also look good

timg236:
Merging now. 

Packaging note. In Bookworm raspi-config has been updated to detect the firmware directory via "rpi-eeprom-update". For older versions of raspi-config (Pi4 only) the workaround is to add a symlink from /lib/firmware/raspberrypi/bootloader-2711 to /lib/firmware/raspberrypi/bootloader

PR #487, opened 2023-10-18
pieeprom-2023-10-18: 2712: Display autodetect + HAT gpiomap (default + latest)

timg236:
* Add support for HAT gpiomap for improved HAT compatibility.
* Add I2C probe for DSI display auto detect
* Automatically set dtparam=nvme if booted from nvme
* Fix network boot reset issue where only the first attempt works.
* Adding pciex4_reset=0 to config.txt will leave RP1 PCIe enabled when ARM stage is started.
* Prevent HDMI diagnostics being displayed immediately when waking after HALT.
* Update board-name - "Rasberry Pi 5"


timg236:
I've cloned this, tested SD / USB boot and power button behaviour and USB current so it seems good. Merging.

PR #489, opened 2023-10-19
Fix /boot/firmware mount

trejan:
Name changed due to path change


timg236:
@XECDesign

XECDesign:
Thanks again

PR #491, opened 2023-10-21
rpi-eeprom-update: Fix typos in usage text

lurch:


rpi-eeprom-update
```diff
@@ -493,7 +493,7 @@ to be stable.
 default:
 The default bootloader image which supports all current models and hardware
 revisions. If there is an important bug fix or hardware change that could
-affect most uses then the then BOOTLOADER_AUTO_UPDATE_MIN_VERSION is updated
+affect most users then the BOOTLOADER_AUTO_UPDATE_MIN_VERSION is updated
```
timg236, line 496: Looks good but please could you make the commit title more descriptive e.g. convention would be something like

rpi-eeprom-update: Fix typos in help text

The scope makes the history to read and indicates that there is no change in behaviour

PR #492, opened 2023-10-24
rpi-eeprom-update: Use flashrom if available on Raspberry Pi 5

timg236:
On Raspberry Pi 5 there are dedicated pins for the bootloader SPI EEPROM. This makes it possible to do immediate updates via flashrom.

The "current" EEPROM config is the EEPROM config at boot rather than what has just been written to the SPI flash because this is consistent with current behaviour.

A potential future improvement would be to add flags to rpi-eeprom-config etc to read from SPI flash rather than nvmem.

TODO: Flashrom needs to set the update timestamp at at the end of the EEPROM image.


timg236:
@XECDesign @peterharperuk FYI, I'm considering making something like this the default on Raspberry Pi 5.

CM4 users might find this useful (rather than rolling their own version) but would need to explicitly enable it.

rpi-eeprom-update
```diff
@@ -237,7 +241,25 @@ applyRecoveryUpdate()
...
+         # Success - remove fallback fails
+         removePreviousUpdates
+         echo "UPDATE SUCCESSFUL"
+      else
+         # Leave the recovery files in case the EEPROM has been partially updated
+         cat "${FLASHROM_LOG}"
+         die "UPDATE FAILED"
+      fi
+      return
+   elif [ "${RPI_EEPROM_SELF_UPDATE}" != "1" ]; then
```
peterharperuk, line 262: Won't "${RPI_EEPROM_SELF_UPDATE}" != "1" always be true?

rpi-eeprom-update
```diff
@@ -237,7 +241,25 @@ applyRecoveryUpdate()
...
+      echo
+      FLASHROM_LOG="$(mktemp)"
+      if flashrom -p linux_spi:dev=${SPIDEV},spispeed=16000 -w "${BOOTFS}/pieeprom.upd" > "${FLASHROM_LOG}"; then
+         # Success - remove fallback fails
+         removePreviousUpdates
+         echo "UPDATE SUCCESSFUL"
+      else
+         # Leave the recovery files in case the EEPROM has been partially updated
+         cat "${FLASHROM_LOG}"
+         die "UPDATE FAILED"
```
peterharperuk, line 259: Is die a bash command?

peterharperuk: Maybe it's worth an echo to say "Update failed, reboot to retry" or something

timg236: Possibly, but the script is #!/bin/sh and die is a function

peterharperuk: ah yes - didn't spot that

rpi-eeprom-update
```diff
@@ -237,7 +241,25 @@ applyRecoveryUpdate()
...
+   # For immediate updates via flash the recovery.bin update is created and then discarded if the
+   # flashrom update was successful. For SD boot (most common) this provides a rollback in the event
+   # of power loss.
+   if [ "${RPI_EEPROM_USE_FLASHROM}" = 1 ]; then
+      echo
+      echo "UPDATING bootloader. Do not disconnect the power until the update is complete"
+      echo
+      FLASHROM_LOG="$(mktemp)"
+      if flashrom -p linux_spi:dev=${SPIDEV},spispeed=16000 -w "${BOOTFS}/pieeprom.upd" > "${FLASHROM_LOG}"; then
+         # Success - remove fallback fails
```
lurch, line 253: Is "fallback fails" supposed to be "fallback files" ?

rpi-eeprom-update
timg236: It defaults to zero but could be forced to 1 via the /etc/defaults file

pelwell: I'm guessing the point is more that this line could be replaced with `else`.

peterharperuk: ^

pelwell: No, either it's changed or (more likely) I misread it - the `if` is about `RPI_EEPROM_USE_FLASHROM`.

rpi-eeprom-update
timg236: changed text

timg236:
After some fun with the --edit mode I think this is now updating the timestamps correctly. I ended up modifying rpi-eeprom-config to have a mode where it can set just the timestamp. This seems cleaner than devious usage of od / awk to hack binaries.

Needs more testing on Pi4 / CM4

rpi-eeprom-update
```diff
@@ -169,7 +173,14 @@ prepareImage()
    if [ "${OVERWRITE_CONFIG}" = 0 ]; then
       "${script_dir}/rpi-eeprom-config" \
          --out "${TMP_EEPROM_IMAGE}" \
-         --config "${NEW_EEPROM_CONFIG}" "${BOOTLOADER_UPDATE_IMAGE}"
+         --config "${NEW_EEPROM_CONFIG}" \
+         --timestamp "$(date -u +%s)"
```
popcornmix, line 177: missing `\` ?

timg236: yes, will fix

timg236:
Update rpi-eeprom-config to improve the UX for the potentially long-running flashrom update command

timg236:
This seems safe so sharing for others to try

PR #498, opened 2023-11-21
pieeprom-2023-11-20: 2712: Auto-detect support for PCIe expansion HAT… (default + latest)

timg236:
* Add autodetect support for PCIe expansion HATs
* Add PCIE_PROBE=1 to the EEPROM config for custom PCIe exapansion designs that do not support the upcoming HAT spec. This gives similar behaviour to CM4 where PCIe x1 is enumerated to discover NVMe devices.
* Fix loading of multiple initramfs images that are not 32-bit aligned sizes https://github.com/raspberrypi/firmware/issues/1843
* Kernel load performance improvement - remove a memcpy


timg236:
@peterharperuk @pelwell

PR #500, opened 2023-11-30
rpi-eeprom: Disable the dpkg checksum validation by default

timg236:
Enforcing package checksum validation makes it difficult to install new EEPROM binaries via rpi-update causing extra churn of APT releases.

Drop this check by default because random corruption of these files either via software or users never actually happens.


timg236:
@peterharperuk @popcornmix

peterharperuk:
makes sense to me

lurch:
If checksums are now ignored by default, should the "-i Ignore package checksums - for rpi-eeprom developers." part of the usage-message be tweaked?

timg236:
> If checksums are now ignored by default, should the "-i Ignore package checksums - for rpi-eeprom developers." part of the usage-message be tweaked?

I've dropped it from the usage since it's effectively redundant unless overridden via the defaults file. The arg is left there to avoid breaking scripts.

PR #509, opened 2023-12-18
Fix typo in warning message

lurch:


pelwell:
I'd say that's stretching the definition of a typo, but we'll take it.

Issue #514, opened 2024-01-05
Raspberry Pi 5 bootloader sometimes treats SFN (only) directory entries as LFNs


timg236:
### Describe the bug

Overlays fail to load on Ubuntu Server 23.10 because the 'overlays' directory entry has no-corresponding LFN entry. This is legal in FAT but the bootloader doesn't always handle this and sometimes depending on the sequencing of state-machine calls the SFN is treated as an LFN which then fail the string comparison part of the match.

### Steps to reproduce the behaviour

See https://forums.raspberrypi.com/viewtopic.php?t=361397#p2176694

### Device (s)

Raspberry Pi 5

### Bootloader configuration.

Any

### System

 Ubuntu Server 23.10

timg236:
Fix candidate here
[rpi-eeprom-recovery.zip](https://github.com/raspberrypi/rpi-eeprom/files/13844132/rpi-eeprom-recovery.zip)

PR #515, opened 2024-01-06
pieeprom-2024-01-05: 2712: Fix handling of FAT files without LFNs.

timg236:
* Fix issues with SFN entries sometimes being treated as LFNs see https://github.com/raspberrypi/rpi-eeprom/issues/514
* Add a dedicated message for "M.2 HAT" not being found instead of the generic 'unsupported boot order' message when NVMe boot is skipped.


timg236:
@peterharperuk @pelwell 

Pushing now so that this is visible to rpi-update users before releasing to all users as an automatic update.

PR #516, opened 2024-01-07
Updated documentation links

trejan:
Documentation anchor IDs had been changed as they aren't Pi 4 specific now.


pelwell:
Thanks!

pelwell:
Not any more, it doesn't.

PR #528, opened 2024-01-24
2024-01-24: NVMe boot fix for WD NVMe (latest)

peterharperuk:
* Add a workaround for an issue seen when booting with WD Blue SN550 NVMe SSD


peterharperuk:
@timg236 Can you just check this look ok? I've tested this commit

timg236 (approved):
Not tested but looks sensible.

PR #537, opened 2024-02-04
Print warning when flashrom is not found, instead of silently ignorin…

### LINKED ISSUE #536: `rpi-eeprom-update` silently ignores `RPI_EEPROM_USE_FLASHROM` when `flashrom` is not found
### Describe the bug

As in the title, `rpi-eeprom-update` silently ignores `RPI_EEPROM_USE_FLASHROM` if `flashrom` is not found.
It would be kind to at least display some warning when the user intentionally sets `RPI_EEPROM_USE_FLASHROM=1`.
I will submit a pull request.

### Steps to reproduce the behaviour

1. Set `RPI_EEPROM_USE_FLASHROM=1` in `/etc/default/rpi-eeprom-update`.
2. Run `rpi-eeprom-update` without installing `flashrom`.

### Device (s)

Raspberry Pi 5

### Bootloader configuration.

.

### System

_No response_

### Bootloader logs

_No response_

### USB boot

_No response_

### NVMe boot

_No response_

### Network (TFTP boot)

_No response_

mocknen:
Fix #536

### DIFF
--- a/rpi-eeprom-update
+++ b/rpi-eeprom-update
@@ -404,6 +404,10 @@ checkDependencies() {
    # Default to off - in the future Raspberry Pi 5 may default to using flashrom if
    # flashrom is available.
    [ -z "${RPI_EEPROM_USE_FLASHROM}" ] && RPI_EEPROM_USE_FLASHROM=0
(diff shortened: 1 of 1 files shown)


timg236:
I was going to ignore this until flashrom support is rolled out but since you already have a patch merging. Thanks!

PR #550, opened 2024-03-01
Add new cacertder option

peterharperuk:
Allows you to add a custom ca cert to an image. Needs an update to the image format to add a space for the new file


timg236:
Looks good, might be worth noting in the commit that this option is only relevant for newer (unreleased) bootloader images
that support custom CA certs and reserve the space in the flash image.

peterharperuk:
Rebased and updated commit text

PR #555, opened 2024-03-27
Secure boot tools

timg236:
* Improve help documentation for rpi-eeprom-digest to make it clearer that this is a low level tool for signing binary files and that the wrapper script must be used for creating a secure-boot SPI flash image.
* Add the 2712 code signing tools to support secure-boot firmware (once released)


timg236:
@peterharperuk @tdewey-rpi

PR #561, opened 2024-04-05
Fix shellcheck issues

Siecje:


pelwell:
I support the first change (removing the dollar signs in arithmetic), but the rest seem unnecessary:
1. Show me a situation where adding the quotes to this script does anything helpful except preventing shellcheck from complaining that they aren't present. Adding the quotes makes it look like spaces might be valid, but the arithmetic says otherwise.
2. Doesn't having non-ASCII in a key mean you have a problem?
3. Show me an echo implementation where the `-n` flag doesn't suppress the newline.

timg236:
This must work on busybox shell.

TBH Keeping spellcheck happy is a non-goal

pelwell:
3. If `${key}` contains a newline (or anything except the hexadecimal data) the wc check will fail, making the shell differences moot.

PR #565, opened 2024-04-16
pieeprom-2024-04-16: 2711: Update Pi4/CM4/Pi400 to using mainline code

timg236:


timg236:
@popcornmix @pelwell @timg236 @peterharperuk @XECDesign 

I've been using this for a while and it seems stable (famous last words!). I plan to release this via rpi-update and leave it there for a bit to check for regressions. It should make it easier to support secure-boot and provisioning.

PR #566, opened 2024-04-17
rpi-eeprom-config - Fix handling of bootcode arg

timg236:
Update to allow all file replacement arguments to be passed in a single call.

Also, add debug flags and update the instructions for installing pycryptodome


timg236:
@tdewey-rpi - Please could check if this resolves the problem that you were seeing?

lurch:
Looking at the code, it seems like some options are ignored if other options aren't set? _Perhaps_ it's worth checking for these invalid combinations (e.g. specifying `--digest` without also specifying `--config`) and printing out a suitable error-message, in order to prevent user-confusion? :shrug:

timg236:
All the signing stuff is wrapped so I think it's ok

PR #569, opened 2024-04-18
pieeprom-2024-04-18: 2712: Update RP1 firmware to extend PCIe L1 entry timeout to 32 us (latest)

timg236:
* Extend PCIe L1 entry timeout to 32us Fix xhci soft reset on link-down Set useful xhci compatibility bits in GUCTL 
* See https://github.com/raspberrypi/firmware/issues/1877


timg236:
@P33M

PR #572, opened 2024-04-20
pieeprom-2024-04-20: 2712: Fix SDRAM default refresh timings

timg236:


timg236:
@popcornmix @XECDesign

PR #604, opened 2024-09-11
pieeprom-2024-10-09: 2712: Promote to default release (automatic update)

timg236:


timg236:
@XECDesign @peterharperuk

PR #624, opened 2024-11-04
Archive older firmware releases

timg236:
Requires https://github.com/RPi-Distro/raspi-config/pull/260


peterharperuk:
Looks ok, but wasn't the last 2711 default release pieeprom-2024-04-15.bin?
I assume vl805-000138c0.bin is the correct latest vl805 version.

timg236:
> Looks ok, but wasn't the last 2711 default release pieeprom-2024-04-15.bin? I assume vl805-000138c0.bin is the correct latest vl805 version.

pieeprom-2023-01-11.bin is the version being programmed at Sony so I left that there. Maybe after the boot-menu stuff is mainline we can push out an automatic update on 2711

PR #627, opened 2024-11-06
rpi-eeprom-digest: honour SOURCE_DATE_EPOCH

Villemoes:
If used in a build environment that sets SOURCE_DATE_EPOCH, we should honour that in order to generate reproducible binaries.

See https://reproducible-builds.org/specs/source-date-epoch/ .

### DIFF
--- a/rpi-eeprom-digest
+++ b/rpi-eeprom-digest
@@ -92,7 +92,11 @@ writeSig() {
    sha256sum "${IMAGE}" | awk '{print $1}' > "${OUTPUT}"
 
    # Include the update-timestamp
-   echo "ts: $(date -u +%s)" >> "${OUTPUT}"
+   if [ -n "${SOURCE_DATE_EPOCH}" ] ; then
+       echo "ts: ${SOURCE_DATE_EPOCH}" >> "${OUTPUT}"
+   else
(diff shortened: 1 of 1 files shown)


timg236:
Please can you elaborate on what's being fixed here?

Villemoes:
I'm trying to make our builds binary reproducible. And yes, there might very well be cases where one doesn't set SOURCE_DATE_EPOCH, and my patch doesn't change anything in such an environment.

But when SOURCE_DATE_EPOCH is set, typically because this is called from some meta-build system, the user/developer/distro/... very much has expressed a desire to generate reproducible builds (https://reproducible-builds.org/), and to that end it is expected that all tools involved follow https://reproducible-builds.org/specs/source-date-epoch/ , in particular "Build processes MUST use this variable for embedded timestamps in place of the "current" date and time. "

timg236:
Ok, I think parameterizing the date is ok and I can see the point about reproducible builds in these systems. Some care must be taken here because a prebaked old timestamp will result in the bootloader skipping the update if it has a newer version from some other route.
Merging on the proviso that this feature is intended for users who actively manage a fleet of Pis and control when updates are applied.

PR #634, opened 2024-11-21
pieeprom: 2024-11-12-2712: Promote 2024-11-12 to default release (default)

timg236:
* Promote 2024-11-12 to the default release and archive older versions.


timg236:
@peterharperuk @popcornmix Push this release to more users - includes the initial turbo settings and boot-menu so give it a few days before making this an automatic update.

PR #639, opened 2024-12-09
Retry flashrom on failure

timg236:
There's been no confirmed reports of flashrom failing but it seems to make sense to retry if it did fail rather than leaving the SPI flash in a potentially broken state.


rpi-eeprom-update
```diff
@@ -189,6 +189,30 @@ prepareImage()
    fi
 }
 
+runFlashrom()
+{
+   retries=5
```
peterharperuk, line 194: Mr pedantic in me would say that this should be named "attempts"

rpi-eeprom-update
```diff
@@ -189,6 +189,30 @@ prepareImage()
...
+      if flashrom -p linux_spi:dev=${SPIDEV},spispeed=${freq} -w "${BOOTFS}/pieeprom.upd" >> "${FLASHROM_LOG}"; then
+         echo "Verifying update"
+         if flashrom -p linux_spi:dev=${SPIDEV},spispeed=${freq} -v "${BOOTFS}/pieeprom.upd" >> "${FLASHROM_LOG}"; then
+            echo "VERIFY: SUCCESS"
+            return 0
+         else
+            echo "VERIFY: FAIL"
+         fi
+      fi
+      echo "Failed to update bootloader SPI flash using flashrom. Retrying."
```
peterharperuk, line 209: Do we want to avoid printing this message on the last iteration?

rpi-eeprom-update
timg236: Already then :)

rpi-eeprom-update
timg236: Yes will change

rpi-eeprom-update
lurch: Just out of curiosity: if it fails on the 3rd attempt is it any more likely to work on the 4th or 5th attempt?

pelwell: We can't know that, but if the alternative is a unit that doesn't boot, how hard would you like us to try?

lurch: Sorry, I was only being curious, I didn't mean to sound rude.

pelwell (approved):

peterharperuk (approved):

PR #642, opened 2024-12-12
Make rpi-eeprom-update and rpi-eeprom-config update prompts clearer

timg236:


pelwell (commented):
There's only one 's' in 'misleading', but otherwise LGTM.

PR #645, opened 2024-12-19
pieeprom-2024-12-19: 2712: Disable PWM fan at shutdown (POWER_OFF_ON_HALT=0) (latest)

timg236:
* Disable fan PWM before shutdown Drive the RP1 fan PWM GPIO high before entering the VPU sleep (POWER_OFF_ON_HALT=0) to stop the fan spinning.
* Disable fan PWM GPIO between RP1 init and fan probe Drive fan PWM GPIO high during early boot to disable the fan until it is probed during the device-tree setup stage. This stops the spinning at max rpm during network-install.
* arm_dt: enable_uart defaults to 0 on 2712 The default value of enable_uart on 2712 is 0, regardless of the presence of the debug UART cable, so guarantee that the default is always set correctly.


jpkelly:
rpi-update says I have the latest firmware but the fan still goes to 100% when shutdown

ulrgaer:
I have the same issue with a cm5 module. Fang goes to 100% after shutdown. Power support via POE. 
Used eeprom:  Mi 8. Jan 17:52:48 UTC 2025 (1736358768)

pelwell:
Ah - you can't respond, but unless POWER_OFF_ON_HALT in the EEPROM configuration is set to 0 then the fan will always spin at shutdown - that's just how the hardware works.

PR #650, opened 2025-01-08
pieeprom-2025-01-08: 2712: Update SDRAM timings for BCM2712D0 producs

timg236:
* Update SDRAM timings for BCM2712D0 products.


timg236:
@popcornmix

Issue #675, opened 2025-03-06
Experimental 2712 DDR init improvements


timg236:
### Describe the bug

Placeholder issue to share experimental firmware that improves DDR init on some marginal boards that encounter intermittent 8-flashes (SDRAM error) 

### Steps to reproduce the behaviour

Flash the attached rpi-eeprom-recovery.zip to a SPARE SD card using the RPi Imager.

Boot the Pi5 with this bootloader recovery image - wait for green HDMI screen / steady flashing pattern.

 If this works then attempt to boot the regular OS image

### Device (s)

Raspberry Pi 5

### Bootloader configuration.

Any

timg236:
[rpi-eeprom-recovery-sdram-init-2025-03-06.zip](https://github.com/user-attachments/files/19111749/rpi-eeprom-recovery-sdram-init-2025-03-06.zip)

timg236:
Added release candidate - simplified version of original patch

[rpi-eeprom-recovery-sdram-init-2025-03-08.zip](https://github.com/user-attachments/files/19145172/rpi-eeprom-recovery-sdram-init-2025-03-08.zip)

eXtorian:
> Added release candidate - simplified version of original patch
> 
> [rpi-eeprom-recovery-sdram-init-2025-03-08.zip](https://github.com/user-attachments/files/19145172/rpi-eeprom-recovery-sdram-init-2025-03-08.zip)

Been having this issue for a week or two (8 fast flashing green lights on power on) on a Pi 5. I want to say it first appeared after doing a "sudo rpi-update", but it's hard to tell as it was a brand new Pi 5.

Rebooting by "sudo shutdown now -r" would cause the issue to appear maybe 50% of the time. Power cycling one or more times would eventually result in an almost normal boot (it would boot into a red/grey screen for a few seconds before proceeding normally). It would also occur via "sudo shutdown now -h" and power cycling.

Anyway, I installed the simplified version of the patch, and so far I've rebooted the Pi 5 about 20 times perfectly normally.

Thank you for your work on this.

holzgeist:
Same here! I had the same issues, this patch (resp the new version that was released 2h ago) works fine

timg236:
Thanks for the test reports. The firmware update is in rpi-update. It should get into apt soon barring any reports of regressions

timg236:
These fixes are in the default and RPi Imager bootloader releases. Closing as completed.

omartio:
[...]

PR #676, opened 2025-03-10
pieeprom-2025-03-10: 2712: Add [boot_partition] filter plus SDRAM init fixes (latest)

timg236:
* Update SDRAM init timings to intermittent 8-flash SDRAM init errors on some boards. 
See: https://github.com/raspberrypi/rpi-eeprom/issues/67
* config: Fix missing initialisation of selected_expr to 1 in config.txt Without an [all] section the new expression filter might default to false. This impacts the bootloader early parsing of config.txt for things like boot_ramdisk rather than the later config.txt pass for device-tree parsing.
* config_loader: Add support [boot_partition=N] as an expression filter The boot_partition tests whether the partition number N matches the number that the system is booting from. This expression is only supported in config.txt and is designed to make it easier to have common boot.img ramdisks in an A/B system where the conditional loads a different cmdline.txt file depending on which partition boot.img is loaded from.


lurch:
Presumably the "See #67" was _supposed_ to be "See #675" ?

timg236:
Yep, I've fixed up the PR link

PR #677, opened 2025-03-11
2712: Promote 2025-10-03 to the default release and archive old releases from the latest sub-directory

timg236:
Bump the default release and archive any 'latest' images that were never promoted to default.


lurch:
`2025-10-03` ?? It's not October yet... :laughing:

PR #740, opened 2025-08-20
rpi-otp-private-key: Describe how to store an ECDSA P-256 private key

timg236:
The Raspberry Pi firmware cryptography service requires a valid ECDSA P-256 key instead of a plain random number. Update the usage instructions for key-provisioning to use this key type as the example.


tools/rpi-otp-private-key
```diff
@@ -134,7 +155,8 @@ if [ $(((0x$BOARD_INFO >> 12) & 15)) = 3 ]; then
 elif [ $(((0x$BOARD_INFO >> 12) & 15)) = 4 ]; then
     MAX_ROW_COUNT=16
 else
-    die "Chip not supported"
+   echo "WARNING: Secure-boot is only supported on Pi4 and newer models"
```
peterharperuk, line 158: Indent is a bit off here? Is there a reason why we allow this to work for unsupported models now?

tools/rpi-otp-private-key
```diff
@@ -30,11 +31,31 @@ usage() {
...
-   to any user in the 'video' group via vcmailbox. Therefore this functionality is only suitable for key
-   storage if the OS has already been restricted using the signed boot functionality.
+   Key generation:
+   The Raspberry Pi firmware cryptography services requires that the device unique private key is
+   a valid ECDSA with P-256 curve key. Due to limited OTP space only the raw private key component (d)
+   is stored in OTP.
+
+   Example key generation and provisioning.
+
+   # Generate the new private-key - remember to create save this to a secure, off-device location!
```
lurch, line 43:
```suggestion
   # Generate the new private-key - remember to save this to a secure, off-device location!
```

tools/rpi-otp-private-key
timg236: done

Issue #790, opened 2025-12-15
shell_cmd should use monotonic time for timeouts, not wall clock time


davidje13:
### Describe the bug

The timeout in `shell_cmd` is based on `time.time()`, which uses wall-clock time:

https://github.com/raspberrypi/rpi-eeprom/blob/c77d47cbecf813ed7da668be516c15842de7d552/rpi-eeprom-config#L120-L125

This means that if the clock is synced while a command is running, it may time out prematurely. I have observed this when attempting to run `raspi-config` commands during a cloud init install:

```
Nov 24 02:07:42 pi5 cloud-init[668]: [snip]
Dec 13 12:22:13 pi5 systemd-timesyncd[416]: Contacted time server 178.62.68.79:123 (2.debian.pool.ntp.org).
Dec 13 12:22:13 pi5 systemd-timesyncd[416]: Initial clock synchronization to Sat 2025-12-13 12:22:13.715210 GMT.
Dec 13 12:22:15 pi5 cloud-init[668]: ERROR: rpi-eeprom-update -d -i -f /tmp/tmp20d6wcmb/pieeprom.upd timeout
```

This is an extreme example of the clock changing, but easy to replicate (by using `raspi-config` commands in `runcmd`). Similar issues could occur any time the system clock is synced with an external source.

Using any of `time.CLOCK_BOOTTIME`, `time.CLOCK_MONOTONIC`, or `time.CLOCK_UPTIME` instead should fix the issue. There may be other places which also need updating to fix similar issues, but I haven't checked through the code.

### Steps to reproduce the behaviour

Run raspi-config (or another eeprom-writing command), for example `sudo raspi-config nonint do_boot_order B2` and change the system time to at least 3 minutes into the future while it is running (observed via NTP above, but should also be possible by manual changes using e.g. `sudo date -s '1 JAN 2030 00:00:00'`)

### Device (s)

Raspberry Pi 5

### Bootloader configuration.

n/a

### System

```
$ cat /etc/rpi-issue
Raspberry Pi reference 2025-11-24
Generated using pi-gen, https://github.com/RPi-Distro/pi-gen, 6e49e5317491c159c114681ddb4ac0623d855611, stage2

$ vcgencmd bootloader_version
2025/11/05 17:37:18
version 57db150d63864d47e6c9071f6b086a5401eb4e92 (release)
timestamp 1762364238
update-time 1763950067
capabilities 0x0000007f

$ vcgencmd version
[...]

PR #791, opened 2025-12-16
rpi-eeprom-config: Use time.monotonic for timeouts instead of wall clock

timg236:
See: https://github.com/raspberrypi/rpi-eeprom/issues/790


lurch (approved):
LGTM. https://docs.python.org/3/library/time.html#time.monotonic says that it was "Added in version 3.3." but I guess that shouldn't cause a problem for anyone!

PR #824, opened 2026-04-27
tools: Add versions table

timg236:
For convenience, add a versions table detailing the firmware git hash, version (date), build timestamp and current release type for every bootloader file.
The table can be regenerated by running tools/update-versions-txt


tools/update-versions-txt
```diff
@@ -0,0 +1,106 @@
...
+release_priority() {
+    case "$1" in
+        default) echo 1 ;;
+        latest)  echo 2 ;;
+        old)     echo 3 ;;
+        *)       echo 9 ;;
+    esac
+}
+
+write_versions_md() {
```
lurch, line 45: `write_versions_txt` ? :wink:

PR #827, opened 2026-05-07
rpi-eeprom-update: Add flashrom check for unknown flash

mathewb64:
Flashrom shouldn't be used if it doesn't know the flash chip from it's ID

### DIFF
--- a/rpi-eeprom-update
+++ b/rpi-eeprom-update
@@ -293,7 +293,6 @@ applyRecoveryUpdate()
       echo "If a problem occurs then the Raspberry Pi Imager may be used to create"
       echo "a bootloader rescue SD card image which restores the default bootloader image."
       echo
-      FLASHROM_LOG="$(mktemp)"
       if runFlashrom; then
          # Success - remove update files from the boot partition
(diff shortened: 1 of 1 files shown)


rpi-eeprom-update
```diff
@@ -344,12 +343,14 @@ applyUpdate() {
    # Disable flashrom if the SPI device is not found
    if [ "${RPI_EEPROM_USE_FLASHROM}" = 1 ]; then
       flashrom_probe_ok=0
+      FLASHROM_LOG="$(mktemp)"
+
       if ! [ -e "${SPIDEV}" ]; then
          warn "WARNING: SPI device ${SPIDEV} not found. Setting RPI_EEPROM_USE_FLASHROM to 0"
```
timg236: Whilst you are here, is it worth removing the duplicate "Setting RPI_EEPROM_USE_FLASHROM to 0" at line 349 because this will be displayed at line 358 anyway.

mathewb64: yep

timg236 (approved):

PR #829, opened 2026-05-12
rpi-eeprom-update: Add OTP bootloader minimum version checks

martamomotko:


rpi-eeprom-update
```diff
@@ -439,6 +442,16 @@ checkDependencies() {
       chipNotSupported
    fi
 
+   # OTP row 46 (BCM2711)/56 (BCM2712) encodes the minimum bootloader version required by this board
+   [ "${BCM_CHIP}" = "2711" ] && otp_row=46 || otp_row=56
+   hardware_info=$(vcgencmd otp_dump | grep "${otp_row}:" | sed 's/.*://')
```
timg236: This should use the device tree node instead of checking raw otp values

rpi-eeprom-update
```diff
@@ -439,6 +443,15 @@ checkDependencies() {
       chipNotSupported
    fi
 
+   if [ -f "${DT_MIN_BOOT_VER}" ]; then
+      MIN_BOOT_VER=$(printf "%d" "0x$(od "${DT_MIN_BOOT_VER}" -v -An -t x1 | tr -d ' ' )")
```
pelwell: You should probably switch to decimal integer mode, and definitely make it big endian. Compare:
```
pi@philpi5:~$ od -An -td4 --endian=big /proc/device-tree/chosen/phandle
         111
```
and:
```
pi@philpi5:~$ od -An -td4 /proc/device-tree/chosen/phandle
  1862270976
```

martamomotko: @pelwell sure, I will fix. Though I noticed the original od ... -t x1 | tr -d ' ' form is also used on line 375. Should I update all instances for consistency, or just this one?

timg236: Would it be useful to have an internal bash function which is can be called to wrap

pelwell: There is an existing utility, fdtget, which should be able to do this, but it needs access to the dtb/fdt file, and /sys/firmware/fdt is currently only readable by root.
```
$ ls -l /sys/firmware/fdt
-r-------- 1 root root 80534 May 13 16:04 /sys/firmware/fdt
```
However, if run with sudo we get:
```
$ sudo fdtget -t u /sys/firmware/fdt /chosen phandle
111
$ sudo fdtget -t s /sys/firmware/fdt /chosen booatargs
reboot=w coherent_pool=1M 8250.nr_uarts=1 pci=pcie_bus_safe ...
```
Given that /proc/device-tree (and the content of /sys/firmware/devicetree/base, which it links to) is world-readable, I don't see why it couldn't be changed in future.

In the meantime, a bash function that takes a path to a node and a property an integer seems sensible.

pelwell: Thinking about it some more, the `tr` is doing more than just stripping the leading whitespace, it's combining hexadecimal bytes into hexadecimal words, where the order is implicitly big endian. And because the alphabetics appear after the digits in ASCII, the sorting just works - no decimal conversion is required. Feel free to leave it as-is.

timg236 (approved):
Looks good @pelwell is this ok with you ?

pelwell (approved):

PR #835, opened 2026-05-20
pieeprom-2026-05-20: recovery: 2712: reboot order and reboot arg option (latest)

timg236:
* recovery: 2712: reboot order and reboot arg option 
Add support for set_reboot_order and set_reboot_arg1 in recovery config.txt, so that recovery.bin can set them when it has been loaded over rpiboot.


timg236:
@mathewb64 @tdewey-rpi

PR #841, opened 2026-05-26
pieeprom-2026-05-26: 2712: Make Pi 5 use the correct entropy source for kaslr-seed and rng-seed (latest)

timg236:
* Make Pi 5 use the correct entropy source 
The kernel looks to Device Tree for the rng-seed and kaslr-seed values as sources of entropy. On Pi 5 the bootsys code enables the HWRNG, but for historical reasons these two DT properties were not using the resulting random data. Fix that.


timg236:
@pelwell

timg236:
@nbuchwitz

PR #850, opened 2026-06-11
rpi-bootloader-version: vcgencmd bootloader_version replacement

timg236:
Add a new script which reports the bootloader_version using device-tree instead of vcgencmd.

Whilst rpi-eeprom-update has used device-tree in preference to vcgencmd for some years now there are lots of places in documentation which refer to 'vcgencmd bootloader_version.

Once this new script is released the documenation can be updated to use the preferred approach.

A related change will also remove the vcgencmd dependency from rpi-eeprom-update.

Issue #765, opened 2025-11-05
No corresponding baud rate setting for `enable_rp1_uart`


JinShil:
### Describe the bug

The `BOOT_UART` configuration has a corresponding `UART_BAUD` setting.
The `console=` kernel command line argument allows for baud rate configuration.
However, the `enable_rp1_uart` doesn't seem to have a corresponding baud rate setting.  Unless, I've overlooked it.

It's helpful to set a high baud rate for the early logging features of the bootloader and kernel as it can shave a few seconds off of the overall boot times.  But, if each stage of the boot process does not have the same baud rate, you end up with output like this:

<img width="1029" height="1186" alt="Image" src="https://github.com/user-attachments/assets/49784fa8-85cf-41aa-9c2c-0086d4987c90" />

See the forum thread that prompted this issue report at https://forums.raspberrypi.com/viewtopic.php?t=391976

### Steps to reproduce the behaviour

Try to set the baud rate for each stage of the boot process to something other than 115200.

### Device (s)

Other

### Bootloader configuration.

[all]
BOOT_UART=1
UART_BAUD=921600
BOOT_ORDER=0xf641
NET_INSTALL_AT_POWER_ON=1
BOOT_WATCHDOG_TIMEOUT=30
HDMI_DELAY=0
PSU_MAX_CURRENT=5000
USB_MSD_DISCOVER_TIMEOUT=5000

[gpio16=1]
BOOT_ORDER=0xf14

[gpio17=1]
BOOT_ORDER=0xf146
PCIE_PROBE=1

timg236:
You can already configure the baud rate for the RP1 UART in the Linux kernel via the command line.

There's no baud rate control for RP1 UART in the bootloader, it doesn't really make sense to try and make this go fast because there is a high overhead (per byte) in talking to RP1 UART in the bootloader since the VPU can't access it directly. Therefore, for boot optimisation just turn off enable_rp1_uart=1

I'll leave it open as low priority feature request.

PR #852, opened 2026-06-17
pieeprom-2026-06-17: 2712: rpi-fw-crypto fine-grained locking (latest)

timg236:
* rpi-fw-crypto fine-grained locking 
Crypto operations can be individually locked per key until after reboot using rpi-fw-crypto set-key-status. Reading, signing, hmac, setting key usage and generating a key can each individually be locked. Setting key usage and generating a key can be locked with lock_device_key_write=1 in config.txt.
* Use UTC for BUILD_DATE and BUILD_TIME 
Expand BUILD_TIMESTAMP using (date -u) for the human readable date / timestamp strings. See: https://github.com/raspberrypi/rpi-eeprom/pull/850
* fix wrap issue with platform_stc64 
It is unsafe to read lo and hi registers separately around the wrap point (every 71m)
* gencmd: Disable pmicrd and pmicwr if secure-boot is enabled
* arm_dt: Avoid incompatible overlay memory leak 
Overlay loading is now aborted early if the overlay map says that it isn't compatible. Unfortunately that error path leaked the memory used to hold any parameters passed to the overlay. Plug the leak.
* Stop the heartbeat with the watchdog 
There is no reason to keep the watchdog heartbeat going when the watchdog is stopped. Ensure the heartbeat is also stopped. See: https://github.com/raspberrypi/firmware/issues/2023
* arm_dt: Defer overlay_map loading until needed 
There is no need to load the overlay map if overlays are not being used. Defer the loading of overlay_map until it is actually needed, saving a few tens of milliseconds.
* arm_dt: Drop an overlay if remapping fails 
If the process of overlay name remapping explicitly fails by returning NULL, don't proceed to apply the overlay anyway - it has been rejected.
* Avoid an unnecessary relocation ARM64 kernels 
Include a load address in the header, but others don't. For those cases, treat an explicit kernel_address setting as gospel, potentially avoiding an unnecessary relocation.
* Set RP1 UART baud to the value configured in eeprom config 
See: https://github.com/raspberrypi/rpi-eeprom/issues/765
* arm-loader: Restrict SET_VOLTAGE to core-voltage on Pi4 and newer. 
With LPDDR4 the SDRAM is initialised by Broadcom's DPFE firmware which does PHY training. Attempting to adjust the SDRAM voltage independently of this will just make the system less stable so switch off this legacy behavior on Pi4 and newer.


timg236:
@pelwell

PR #858, opened 2026-07-17
imager: build rpiboot GPIO image without sudo

FraFrieFa:
## Summary

`imager/make-pi4-rpiboot-gpio-sd` currently requires root because it uses
`kpartx`, `/dev/mapper`, `mkfs.fat` on a mapped block device, and `mount` to
populate the FAT partition.

This changes the script to build the same raw SD card image entirely as regular
files without sudo:

- create the DOS/MBR image with `sfdisk`
- create a partition-sized FAT32 image with `mkfs.fat`
- copy `recovery.bin` and `config.txt` with `mcopy`
- write the FAT image into the raw SD image at the existing partition offset
- preserve the existing `.zip` output containing the raw `.img`

It also adds README documentation for the helper, including usage, image format,
valid BCM GPIO values, and Debian/Ubuntu dependencies.

## Testing

- Ran `git diff --check`
- Verified usage output for missing arguments
- Verified invalid GPIO handling
- Built an image as a non-root user
- Confirmed the generated zip contains the raw `.img`

### DIFF
--- a/imager/make-pi4-rpiboot-gpio-sd
+++ b/imager/make-pi4-rpiboot-gpio-sd
@@ -53,6 +53,8 @@ build_image()
    img="${img}.img"
    partition_start=2048
    partition_size=524288
+   temp_disk_img="temp.img"
+   temp_partition_img="fat.img"
 
    TMP_DIR="$(mktemp -d)"
    (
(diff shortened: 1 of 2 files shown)


FraFrieFa:
Note: if this is accepted, `raspberrypi/usbboot` has a small README section for this helper that should be updated as a follow-up to remove the old `sudo`/`kpartx` instructions and list the new non-root dependencies.

imager/make-pi4-rpiboot-gpio-sd
```diff
@@ -68,46 +68,32 @@ EOF
       cat config.txt
       cd "${TMP_DIR}"
       dd if=/dev/zero bs=1M count=258 of=temp.img > /dev/null 2>&1
```
lurch: I wonder if it's also worth adding e.g.:
```bash
temp_disk_img="temp.img"
temp_partition_img="fat.img"
```
:shrug:

imager/make-pi4-rpiboot-gpio-sd
```diff
@@ -131,4 +117,20 @@ case "${gpio_num}" in
...
+   die "sfdisk not found: Try installing the util-linux package"
+fi
+
+if ! MKFS_FAT=$(command -v mkfs.fat); then
+   die "mkfs.fat not found: Try installing the dosfstools package"
+fi
+
+if ! command -v mcopy > /dev/null; then
+   die "mcopy not found: Try installing the mtools package"
+fi
```
lurch, line 132: Why does this create an explicit variable for the path to the `mkfs.fat` executable, but not do the same for the `mcopy` executable?

FraFrieFa:
Resolved

timg236:
Looks useful, leave it with me to have a play and review.

FraFrieFa:
Reminder: [this Readme](https://github.com/raspberrypi/usbboot/blob/master/Readme.md) still mentions the old sudo path and kpartx. (Probably needs mtools/zip on Ubuntu)

PR #874, opened 2026-09-07
rpi-eeprom-update: Skip updates if the current bootloader version is unknown

timg236:
Add some defensive code to ensure that missing / invalid device tree nodes cannot trigger an automatic firmware update.

Guard readDtInt against an empty property and do not pass an empty string to date, which would otherwise be interpreted as today.

Remove dead code for the original 'launch' version of the Pi4 bootloader.


timg236:
@learmj FYI

PR #875, opened 2026-09-07
rpi-eeprom-update: Firmware downgrades now require the '-y' flag or an interactive response

timg236:
In addition to the version check change the default behaviour to never downgrade the bootloader to an older version without prompting the user unless the '-y' flag is passed.

Also, add defensive code to verify that the update timestamp specified in device-tree is valid and that the node is present e.g. if u-boot is used and device-tree does not match RPI OS expected layout for the bootloader.

### DIFF
--- a/rpi-eeprom-update
+++ b/rpi-eeprom-update
@@ -55,8 +55,6 @@ EXIT_EEPROM_FROZEN=3
 # EXIT_PREVIOUS_UPDATE_FAILED=4
 
 OVERWRITE_CONFIG=0
-# Timestamp for first release which doesn't have a timestamp field
-BOOTLOADER_FIRST_VERSION=1557513636
 EEPROM_SIZE=524288
 # BOOT_VER is checked against the manufacturing version in OTP to verify
 # that this script is new enough to correctly update the bootloader on this board.
@@ -128,7 +126,8 @@ warn() {
 }
 
 readDtInt() {
-   printf "%d" "0x$(od "$1" -v -An -t x1 | tr -d ' ')"
+   hex="$(od "$1" -v -An -t x1 | tr -d ' ')"
+   [ -z "${hex}" ] || printf "%d" "0x${hex}"
 }
 
 getBootloaderConfig() {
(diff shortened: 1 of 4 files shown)


rpi-eeprom-update-default
```diff
@@ -19,3 +19,6 @@ EEPROM_CONFIG_HOOK=
 # Set STRICT_MIN_VER_CHECK to 1 to treat bootloader manufacturing min-version checks as a fatal error
 # Default is currently 0
 #STRICT_MIN_VER_CHECK=1
+
+# Set to 1 to install an older bootloader without confirmation
+#RPI_EEPROM_ALLOW_DOWNGRADE=0
```
pelwell: It's minor nit-pick, but the other commented-out values use non-default values, i.e. uncomment to change the behaviour

timg236: Yep, will fix.

rpi-eeprom-update-default
```diff
@@ -16,8 +16,7 @@ EEPROM_CONFIG_HOOK=
 # AB_EEPROM can be set here to override auto-detection of AB EEPROM support
 #AB_EEPROM=
 
-# Set STRICT_MIN_VER_CHECK to 1 to treat bootloader manufacturing min-version checks as a fatal error
-# Default is currently 0
+# Set to 0 to only warn if the image MFG_VER is older than the board requires
 #STRICT_MIN_VER_CHECK=1
```
pelwell: I guess this should also be flipped now

rpi-eeprom-update
```diff
@@ -485,29 +483,64 @@ applyUpdate() {
...
+
+   # Callers such as rpi-eeprom-config redirect stdout so use the tty
+   if ! ( : < /dev/tty ) 2> /dev/null; then
+      die "Bootloader downgrade refused. Use -y or set RPI_EEPROM_ALLOW_DOWNGRADE=1 in /etc/default/rpi-eeprom-update to allow this."
+   fi
+   current="unknown"
+   [ -z "${BOOTLOADER_CURRENT_VERSION}" ] || current="$(date -u "-d@${BOOTLOADER_CURRENT_VERSION}" +%F)"
+   printf "Install bootloader %s over the current version (%s)? [y/N] " \
+      "$(date -u "-d@${BOOTLOADER_UPDATE_VERSION}" +%F)" "${current}" > /dev/tty
+   read -r answer < /dev/tty
```
popcornmix: No input (e.g. ctrl-D) will not trigger the `die` and will update. `offerUpdatePrompt` does a `|| return 1` in similar code.

rpi-eeprom-update
```diff
@@ -128,7 +128,8 @@ warn() {
 }
 
 readDtInt() {
-   printf "%d" "0x$(od "$1" -v -An -t x1 | tr -d ' ')"
+   hex="$(od "$1" -v -An -t x1 | tr -d ' ')"
+   [ -n "${hex}" ] && printf "%d" "0x${hex}"
```
popcornmix: Calude: 
```
readDtInt (line 130) aborts the script on an empty DT property. [ -n "${hex}" ] && printf ... returns 1, and VAR=$(readDtInt ...) takes that status, so the shell exits — silently, with status 1, which is EXIT_UPDATE_REQUIRED. This defeats the whole
     point of the defensive change in 6cbcf97, and makes the new MIN_BOOT_VER=${MIN_BOOT_VER:-0} line unreachable. Confirmed in dash, bash and busybox. Fix: [ -z "${hex}" ] || printf ....
```

rpi-eeprom-update-default
timg236: Yes that as well and I've updated the help to describe how to set the override on the rpi-eeprom-update cmdline

rpi-eeprom-update
timg236: ctrl-C and ctrl-D both abort when I ran it

rpi-eeprom-update
timg236: should be fixed now

popcornmix (approved):

pelwell (approved):

timg236:
Sorry, I added two more comments. I thought the ctrl-C behaviour from 'rpi-eeprom-config --edit' in the 'do you want to proceed prompt' was a bi t ugly because you got a Python exception backtrace.
Added the exception handler and nicer stderr handling

rpi-eeprom-update
timg236: Checked with manual testing and the higher level checks catch this

PR #882, opened 2026-09-14
rpi-eeprom-update: Track whether there is a pending firmware update

timg236:
Bootloader updates don't take effect until the next watchdog or cold boot. Store a file in /run to track whether the update is still pending. This is useful for A/B updates and for suspend-to-ram because resume would fail if the early bootcode loaded from SPI flash into the L2-cache does not match the verison of firmware already in SDRAM.


timg236:
@pineapplebeech @mathewb64

mathewb64 (approved):
LGTM

